PR #2, opened 2020-02-05
Initial spec work

adlerjohn:
- [ ] Data structures
- [ ] Consensus - block validity
- [ ] Networking spec


specs/data_structures.md
```diff
@@ -29,45 +28,51 @@ Data Structures
...
 ## Data
 
- | name  | type                            | description |
- | ----- | ------------------------------- | ----------- |
- | `txs` | [Transaction](#transaction)`[]` |             |
+Wrapper for transaction data, which is a simple list of [Transaction](#transaction)s.
+
+ | name  | type                            | description           |
+ | ----- | ------------------------------- | --------------------- |
+ | `txs` | [Transaction](#transaction)`[]` | List of transactions. |
```
Liamsi: In the paper and in @musalbas' prototype the LL Transactions are called messages. Maybe we can stick to this in the spec here, too? I don't feel strongly about this but then we can have a clear distinction between Tx on a sidechain and Tx to LL layer1 Txs (aka messages).

adlerjohn: Agreed.

musalbas: To be clear, messages in my prototype are blobs of data. It seems here transactions are referring to transactions in the LazyLedger consensus-critical sidechain. In that case I think they should still be called transactions, and be clarified that these are LazyLedger "main sidechain" transactions (whatever we call that).

specs/data_structures.md
```diff
@@ -29,45 +29,54 @@ Data Structures
...
+| `lastBlockID`        | [BlockID](#blockid)         | Previous block's ID.                                        |
+| `lastCommitRoot`     | [HashDigest](#hashdigest)   | Previous block's Tendermint commit root.                    |
+| `dataRoot`           | [HashDigest](#hashdigest)   | Message data root.                                          |
+| `validatorsRoot`     | [HashDigest](#hashdigest)   | Validator set root for this block.                          |
+| `nextValidatorsHash` | [HashDigest](#hashdigest)   | Root of the next block's validator set.                     |
+| `consensusHash`      | [HashDigest](#hashdigest)   | Consensus parameters for this block.                        |
+| `appHash`            | [HashDigest](#hashdigest)   |                                                             |
+| `lastResultsHash`    | [HashDigest](#hashdigest)   |                                                             |
+| `evidenceRoot`       | [HashDigest](#hashdigest)   | Evidence data root.                                         |
+| `codingRoots`        | [CodingRoots](#codingroots) | Commitments (i.e. Merkle roots) of the erasure-coded block. |
```
musalbas: What's the different between coding roots and data root? Aren't these the same?

specs/data_structures.md
```diff
@@ -29,45 +29,54 @@ Data Structures
...
+| `time`               | [Time](#time)               | Timestamp of this block.                                    |
+| `lastBlockID`        | [BlockID](#blockid)         | Previous block's ID.                                        |
+| `lastCommitRoot`     | [HashDigest](#hashdigest)   | Previous block's Tendermint commit root.                    |
+| `dataRoot`           | [HashDigest](#hashdigest)   | Message data root.                                          |
+| `validatorsRoot`     | [HashDigest](#hashdigest)   | Validator set root for this block.                          |
+| `nextValidatorsHash` | [HashDigest](#hashdigest)   | Root of the next block's validator set.                     |
+| `consensusHash`      | [HashDigest](#hashdigest)   | Consensus parameters for this block.                        |
+| `appHash`            | [HashDigest](#hashdigest)   |                                                             |
+| `lastResultsHash`    | [HashDigest](#hashdigest)   |                                                             |
+| `evidenceRoot`       | [HashDigest](#hashdigest)   | Evidence data root.                                         |
```
musalbas: Does the slashing evidence need data availability? If so, they should be under the erasure coded data root.

specs/data_structures.md
```diff
@@ -78,37 +87,36 @@ TODO define a transaction format
...
 
 ## BlockID
 
-| name          | type                            | description |
-| ------------- | ------------------------------- | ----------- |
-| `hash`        | [HashDigest](#hashdigest)       |             |
-| `partsHeader` | [PartSetHeader](#partsetheader) |             |
+The block ID is comprised of two distinct Merkle roots:
+1. The root of the [block header](#header)'s fields, in the order provided in the spec.
+1. The root of the complete [serialized](#serialization) block [split into parts](#todo). This is used at the network layer for securely gossiping parts of blocks.
```
musalbas: What's the difference between the partsHeader and the data root?

specs/data_structures.md
```diff
@@ -78,37 +87,36 @@ TODO define a transaction format
...
+| `partsHeader` | [PartSetHeader](#partsetheader) | Parts header.             |
 
 ## HashDigest
 
 ## Address
 
-## Transaction
+## CodingRoots
+
+## Message
```
musalbas: I'm assuming these are transactions for the LL main sidechain? If so, are these transactions posted directly onto the block, or are they aggregated into a sidechain block like an optimistic rollup?

specs/data_structures.md
adlerjohn: Yes, here messages refers to things that change the consensus-critical state of LazyLedger, aka transactions for normal blockchains. In that context I guess "transactions" would be preferred.

specs/data_structures.md
adlerjohn: The coding roots would be Merkle roots of the erasure-coded block body, while the data root would be the Merkle root of the raw block body (well, just the message data part of it). I think the latter is needed as there's no easy way to prove inclusion of individual transactions (or messages, as it were) without the Merkle root of the raw messages.

specs/data_structures.md
adlerjohn: Actually, maybe not? Evidence is needed to slash validators, which means it affects the validator set, which means it needs to be fully downloaded by both full and light nodes I think.

specs/data_structures.md
adlerjohn: My understanding of the Tendermint design is that it allows blocks to be split into parts that can be securely gossiped in parallel to/from different many different machines. This is an implementation detail really, but apparently an important one. I imagine it might be possible to combine this with the splitting of the erasure coded block in some way (might introduce tight coupling?).

musalbas:
The issue with this is that there is no guarantee that the raw data root 
actually matches the data in the erasure coded root. Therefore there should 
be only one data root.

I don't see why light clients can't verify inclusion proofs in the erasure 
coded root? The tree will still contain raw messages. In the LazyLedger 
prototype I have light client proofs for erasure coded data roots under the 
2D-RS scheme. It's more complicated but still possible.

specs/data_structures.md
adlerjohn: The plan here was for transactions (messages) to be posted individually. Given their consensus-critical nature it shouldn't make any different if they're aggregated or not.

musalbas: They need to be aggregated into state roots so that fraud proofs can be generated however

specs/data_structures.md
adlerjohn: @musalbas says:
> The issue with this is that there is no guarantee that the raw data root
actually matches the data in the erasure coded root. Therefore there should
be only one data root.
>
> I don't see why light clients can't verify inclusion proofs in the erasure
coded root? The tree will still contain raw messages. In the LazyLedger
prototype I have light client proofs for erasure coded data roots under the
2D-RS scheme. It's more complicated but still possible.

Ah, good point, I see now.

specs/data_structures.md
adlerjohn: Yes, that's what the `appHash` field in the block header is going to be for.

musalbas: What about intermediate state roots?

adlerjohn: Couldn't you technically commit to the intermediate state roots and the final state root in a single commitment?

musalbas: No because the intermediate state root need data availability in order to prove fraud, so they need to be included in the data root. Hence why some kind of block aggregation is needed.

adlerjohn: Ah, good point, I see now.

specs/data_structures.md
ebuchman: Right, this is kind of an unfortunate coupling between the blockchain data structure and the consensus reactor's gossip mechanism - it deserves a new issue and consideration: https://github.com/tendermint/spec/issues/78. Also interested how this relates to erasure coded blocks

specs/data_structures.md
```diff
@@ -24,50 +23,63 @@ Data Structures
...
-| `proposerAddress`    | [Address](#address)       |             |
-
-## Data
-
- | name  | type                            | description |
- | ----- | ------------------------------- | ----------- |
- | `txs` | [Transaction](#transaction)`[]` |             |
-
-TODO define a transaction format
+Block header, which is fully downloaded by both full clients and light clients.
```
ebuchman: Note we're considering some changes that might simplify the header: https://github.com/tendermint/tendermint/issues/2530. Would be cool to have your input.

specs/data_structures.md
musalbas: From my understanding the data root is basically an erasure coded partsHeader, because the leafs of the data root are fixed-sized chunks. This can actually be used to build a more efficient gossiping protocol, using information dispersal algorithms. See section 4.3.1 of https://eprint.iacr.org/2018/460.pdf.

Suppose that you're a validator that just made a block. Instead of gossiping plaintext chunks, you can gossip different erasure coded chunks of the block of your neighbors. The data can still be recovered even if some of your neighbors refuse to gossip your chunks, because you can recover the data from only a subset of the chunks. Therefore you do not need to upload as much data to the network in order to distribute your block.

musalbas: By the way, as mentioned previously, we cannot have multiple Merkle roots pointing to the same data. This is because we cannot guarantee that the data in dataRoot is actually the same as the data in the partsHeader, which would defeat the point of data availability proofs.

Liamsi: I've mentioned a very similar idea chatting with @brapse a while ago. I just didn't know about  information dispersal algorithms. It would be awesome if we can implement this as it is needed for LL but could also improve the gossiping mechanism.

specs/data_structures.md
```diff
@@ -161,11 +171,9 @@ https://developers.google.com/protocol-buffers/docs/proto3
...
 
-https://en.wikipedia.org/wiki/SHA-3
+All protocol-level hashing is done using [Keccak-256](https://keccak.team/keccak.html), and not SHA3-256 ([FIPS 202](https://keccak.team/specifications.html#FIPS_202)). This is to enable compatibility with [Ethereum](https://ethereum.org)'s EVM. Keccak-256 outputs a digest that is 256 bits (i.e. 32 bytes) long.
 
-https://godoc.org/golang.org/x/crypto/sha3
-
-https://docs.rs/sha3
+Libraries implementing Keccak-256 are available in Go (https://godoc.org/golang.org/x/crypto/sha3) and Rust (https://docs.rs/sha3).
 
 # Signing
```
Liamsi: Note: for signing consensus data (votes etc), the structs are still binary/proto3 encoded but "canonicalized" and length prefixed:
https://github.com/tendermint/tendermint/blob/aeb6cc475e14c9b4c4455111f6dfb54672507006/types/vote.go#L83-L89

Liamsi: It used to be JSON data but apparently this was too expensive to parse in the evm: https://github.com/tendermint/tendermint/issues/1622

specs/architecture.md
```diff
@@ -0,0 +1,63 @@
...
+
+Cons:
+* Compared to the other proposals, additional commitments are needed in the block header.
+* Due to having two separate environments for validators and fee payments, and LazyLedger not supporting execution at the base layer, only a one-way bridge validator set $\rightarrow$ fee payment is allowed. Transfer to funds backwards can be simulated to a limited extent using atomic swaps between the two environments---HTLCs are pure with respect to state and supporting then with transactions should not add meaningful overhead. Tax and market (i.e. potentially two coins) considerations might come into play.
+
+# Proposal 2: Validator Set and Fee Payments in State
+
+As above, but both the validator set changes and fee payments are transactions that modify state.
```
Liamsi: They modify state in the LL chain, right?

specs/architecture.md
```diff
@@ -0,0 +1,63 @@
...
+
+Pros:
+* Fewer commitments in block headers.
+* Potentially cleaner design, with no pollution in the base layer.
+
+Cons:
+* Requires both non-consensus full nodes and light nodes to additionally download and process Merkle paths to the consensus-critical app's header.
+* As above, only a one-way bridge is supported.
+
+# Proposal 4: Validator Set and Fees in a Consensus-Critical App
```
Liamsi: What is the exact difference from Proposal 3 and 4?

specs/consensus.md
```diff
@@ -0,0 +1,22 @@
...
+# System Parameters
+
+| name                         | type     | value | description                                                      |
+| ---------------------------- | -------- | ----- | ---------------------------------------------------------------- |
+| `SHARE_SIZE`                 | `uint64` | `256` | Size of transaction and message shares, in bytes.                |
+| `SHARE_RESERVED_BYTES`       | `uint64` | `1`   | Bytes reserved at the beginning of each share.                   |
+| `INTERMEDIATE_ROOT_INTERVAL` | `uint64` | `1`   | Maximum number of transactions between intermediate state roots. |
+
+# Leader Selection
+
```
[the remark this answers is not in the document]
Liamsi: Here is a description of the proposer selection in tendermint: https://docs.tendermint.com/master/spec/reactors/consensus/proposer-selection.html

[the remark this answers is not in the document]
Liamsi: Note that from looking at tendermint as a more general library, it would be nicer if users could more easily change the underlying proposer election mechanism. I thought there was an issue about this (and that the proposer election algo should be defined as a pure function of height, round, and last proposer, but couldn't find it). In any case, this could look sth along this lines: https://github.com/celo-org/celo-blockchain/blob/ffa8d5ea42d26cd0636eda197f37b101598c0de4/consensus/istanbul/validator.go#L125:6

This would also be a refactoring that could be contributed back upstream.

specs/data_structures.md
```diff
@@ -78,37 +86,39 @@ TODO define a transaction format
...
-| `hash`        | [HashDigest](#hashdigest)       |             |
-| `partsHeader` | [PartSetHeader](#partsetheader) |             |
+The block ID is comprised of two distinct Merkle roots:
+1. The root of the [block header](#header)'s fields, in the order provided in the spec.
+1. The root of the complete [serialized](#serialization) block [split into parts](#todo). This is used at the network layer for securely gossiping parts of blocks.
+
+| name          | type                            | description               |
+| ------------- | ------------------------------- | ------------------------- |
+| `headerRoot`  | [HashDigest](#hashdigest)       | Root of the block header. |
+| `partsHeader` | [PartSetHeader](#partsetheader) | Parts header.             |
```
Liamsi: My understanding is that we would definitely remove the `partsHeader`. It is only needed by validators to gossip block data, which validators are not obliged to do (it's optional). And as @musalbas suggested this could be replaced by another gossiping mechanism that uses erasure coded data ([here](https://github.com/LazyLedger/lazyledger-specs/pull/2/files#r376971666) and [here](https://github.com/tendermint/spec/issues/78#issuecomment-584052848)).

specs/data_structures.md
```diff
@@ -161,11 +171,9 @@ https://developers.google.com/protocol-buffers/docs/proto3
```
Liamsi: I think the header of this section should be called `# Wire format`. This is only about data serialization over the network. E.g. formats of data signed, or stored can be different (e.g. JSON, sth canonicalized). While the spec needs to define how signed data gets signed, the implementer is free to chose whatever they like for storage.

specs/consensus.md
```diff
@@ -0,0 +1,22 @@
...
+# Fork Choice
+
+# Block Validity
```
Liamsi: While it is straightforward to define _block validity_, we should prioritise specing out how validators are supposed to know _where_ to get the random chunks they require to verify the block is valid! This plays into specing the networking / p2p layer, too.

specs/data_structures.md
```diff
@@ -161,23 +174,88 @@ https://developers.google.com/protocol-buffers/docs/proto3
...
+
+For leaf node of message `m`:
+```C++
+v = h(serialize(m))
+```
+
+For internal node with children `l` and `r`:
+```C++
+v = h(l, r) = h(l.v, r.v)
+```
```
Liamsi: Isn't it essential that leaf nodes vs internal nodes get hashed differently? Usually that happens with a prefix. e.g. https://github.com/tendermint/tendermint/blob/48f073d796ca4d1063b5f02c5d1b35c2c7e77afc/crypto/merkle/hash.go#L8-L21
Some discussions about this here: https://github.com/tendermint/tendermint/issues/1892

Liamsi: Note that the flaghasher in @musalbas prototype also differentiates between leaf (and other) data: https://github.com/LazyLedger/lazyledger-prototype/blob/60faf6af245830ebe25a4963d587e96a467d2eb1/flaghasher.go#L71-L76 
(it additionally has a `codedFlag` if in codec mode). And the leaf hash is computed prefixed with `0x00` (using that flaghasher): https://github.com/LazyLedger/lazyledger-prototype/blob/60faf6af245830ebe25a4963d587e96a467d2eb1/merkletree_exports.go#L20-L22

specs/data_structures.md
```diff
@@ -161,23 +174,88 @@ https://developers.google.com/protocol-buffers/docs/proto3
...
+
+The `namespace_id` field is the namespace ID of the message, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-byte unsigned integer.
+
+Before being hashed, the [message](#message)s are [serialized](#serialization).
+
+For internal node with children `l` and `r`:
+```C++
+n_min = min(l.n_min, r.n_min)
+n_max = max(l.n_max, r.n_max)
+v = h(l, r) = h(l.min, l.max, l.v, r.min, r.max, r.v)
```
Liamsi: Maybe my above comment doesn't matter. As namespacing effectively differentiates between leaf and inner nodes, too.

specs/architecture.md
```diff
@@ -0,0 +1,63 @@
...
+- [Proposal 1: Validator Set in State, Fees in App](#proposal-1-validator-set-in-state-fees-in-app)
+- [Proposal 2: Validator Set and Fee Payments in State](#proposal-2-validator-set-and-fee-payments-in-state)
+- [Proposal 3: Validator Set in a Consensus-Critical App, Fees in App](#proposal-3-validator-set-in-a-consensus-critical-app-fees-in-app)
+- [Proposal 4: Validator Set and Fees in a Consensus-Critical App](#proposal-4-validator-set-and-fees-in-a-consensus-critical-app)
+
+# Common Terms and Expressions
+
+| name              | description                                                                          |
+| ----------------- | ------------------------------------------------------------------------------------ |
+| App (application) | Alternate name for "virtual sidechain."                                              |
```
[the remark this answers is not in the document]
adlerjohn: Yeah agreed that it's somewhat of a naming collision, but for good reason: "apps" here also define how state is modified and state transitions are validated, just not on the main chain. I'll add a more expansive description.

specs/consensus.md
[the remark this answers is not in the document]
adlerjohn: Not sure yet if we want to go the round-robin or randomized (Tenderand?) route. I don't inherently see any major issues with round-robin since the leader is predictable even if randomized.

The consensus protocol used by apps to internally select block proposers for apps (which are then passed to LazyLedger block proposers) is orthogonal, but we should definitely keep it in mind in case there are common design decisions that can be made.

specs/consensus.md
```diff
@@ -0,0 +1,22 @@
...
+| name                         | type     | value | description                                                      |
+| ---------------------------- | -------- | ----- | ---------------------------------------------------------------- |
+| `SHARE_SIZE`                 | `uint64` | `256` | Size of transaction and message shares, in bytes.                |
+| `SHARE_RESERVED_BYTES`       | `uint64` | `1`   | Bytes reserved at the beginning of each share.                   |
+| `INTERMEDIATE_ROOT_INTERVAL` | `uint64` | `1`   | Maximum number of transactions between intermediate state roots. |
+
+# Leader Selection
+
+# Fork Choice
+
```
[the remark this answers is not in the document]
adlerjohn: That's true, given that Tendermint is a BFT consensus protocol it's "fork-free." Majority reorgs can only be dealt with at the social layer anyways.

specs/architecture.md
adlerjohn: Yes.

specs/consensus.md
adlerjohn: Yep. This section should be relatively short, and only cover
1. proper formatting
1. availability (defer the work here to the networking spec)
1. potential state transitions to the validator set

specs/data_structures.md
adlerjohn: Yes, I'm pretty sure that you can't execute the same second-preimage attack here.

specs/data_structures.md
adlerjohn: Hmm I was thinking this section was more defining a canonical serialization format, for use with signing and hashing, etc. Data is transferred over the wire could be different, and should be in a separate networking document.

specs/architecture.md
adlerjohn: In proposal 3, fee payments for data blobs are not in a consensus-critical execution space. In proposal 4 they are.

specs/consensus.md
musalbas: Hmm, maybe we should divide the spec into two parts: 1) the core consensus critical parts of the spec (mesage, tree formats etc), and 2) the non-consensus critical networking protocols.

adlerjohn: Yes, the above discussion had the separation of local consensus rules from networking in mind. The separate networking spec will be in https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-initial_data_structures/specs/networking.md.

adlerjohn:
Most of the core specs are done. Still to do, in future PRs:
- [ ] Consensus rules
- [ ] Transaction and message format
- [ ] Consensus messaging (votes, etc.) format
- [ ] Complete signature spec
- [ ] Networking

specs/consensus.md
```diff
@@ -0,0 +1,52 @@
...
+| `3u` | `M1u` | `10**6` | `1000000` units.    |
+| `4u` | `G1u` | `10**9` | `1000000000` units. |
+
+## Constants
+
+| name                   | type     | value   | unit   | description                                       |
+| ---------------------- | -------- | ------- | ------ | ------------------------------------------------- |
+| `NAMESPACE_ID_BYTES`   | `uint64` | `32`    | `byte` | Size of namespace ID, in bytes.                   |
+| `SHARE_SIZE`           | `uint64` | `256`   | `byte` | Size of transaction and message shares, in bytes. |
+| `SHARE_RESERVED_BYTES` | `uint64` | `1`     | `byte` | Bytes reserved at the beginning of each share.    |
```
Liamsi: We need to add "share" to a terminology section.

specs/data_structures.md
adlerjohn: After some deliberation, putting the evidence data under the data availability root is a good idea because it enables super-light clients that don't actually process validator set changes. Refactoring in #12.

specs/architecture.md
```diff
@@ -0,0 +1,30 @@
...
+
+![fig: Block data structures.](figures/block_data_structures.svg)
+
+LazyLedger has a minimal state: the validator set (account balances, validator status, etc.). Changes to the validator set are done with native _transactions_, distinct from the _messages_ processed by apps. Transactions are signed and must be processed by clients to determine the validator set, while messages are un-signed data blobs that will usually represent an app's block data.
+
+Transactions pay fees similarly to how they would in a normal blockchain (e.g. Bitcoin), and their side effects are restricted to modifying the validator set and their balances. Transactions can additionally pay fees for the inclusion of a message (identified by a hash) in the same block. The validator set is committed to in the block header, and since the entire system state _is_ the validator set, this is the only state commitment needed in the header.
+
+One dideratum that will most likely be included is burning a non-proportional amount of coins for each transaction as a network fee. This provides baseline demand for the native coin: as the chain is used more, more coins must be bought then burned to pay for fees.
```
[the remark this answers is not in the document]
adlerjohn: Fixed in #13.

specs/architecture.md
```diff
@@ -0,0 +1,30 @@
...
+
+| name              | description                                                                                                                                                                      |
+| ----------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| App (application) | Alternate name for "[virtual sidechain](https://arxiv.org/abs/1905.09274)." LazyLedger apps are sidechains that post all their data onto the LazyLedger chain to share security. |
+| Transaction       | Requests that modify the consensus-critical state (validator balances and statuses).                                                                                             |
+| Message           | Requests that are executed by non-consensus-critical apps.                                                                                                                       |
+
+# System Architecture
+
+![fig: Block data structures.](figures/block_data_structures.svg)
```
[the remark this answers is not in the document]
adlerjohn: We still need a single state root in the top-level header, which should replace the validators roots. That's orthogonal to the block data roots, but only a single root is required so that intermediate state roots in the block data can be linked to it for fraud proofs.

specs/architecture.md
```diff
@@ -0,0 +1,30 @@
...
+
+LazyLedger has a minimal state: the validator set (account balances, validator status, etc.). Changes to the validator set are done with native _transactions_, distinct from the _messages_ processed by apps. Transactions are signed and must be processed by clients to determine the validator set, while messages are un-signed data blobs that will usually represent an app's block data.
+
+Transactions pay fees similarly to how they would in a normal blockchain (e.g. Bitcoin), and their side effects are restricted to modifying the validator set and their balances. Transactions can additionally pay fees for the inclusion of a message (identified by a hash) in the same block. The validator set is committed to in the block header, and since the entire system state _is_ the validator set, this is the only state commitment needed in the header.
```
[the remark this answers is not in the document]
adlerjohn: > So to send a message, I would first need to submit a Tx with a reference to the message (which is then included because a fee for it was payed)?

That's the plan with this.

> Is it necessary that inclusion happens in the same block?

I would say so, yes. 1) You can't have messages in a previous block, obviously, as validators can't see the future. 2) If you allow messages to be included in _future_ blocks, you need to have some sort of guarantee they will be processed, which adds overhead and complexity, opens up the door to DoS vectors, etc. Must simpler to just have whoever is assembling the block be required to make sure that any tx that pays a fee for a message also has that message attached in the same block.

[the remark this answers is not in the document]
Liamsi: > Must simpler to just have whoever is assembling the block be required to make sure that any tx that pays a fee for a message also has that message attached in the same block.

What if whoever is assembling the block didn't receive this message? Does the sender of the Tx receive their paid fee back?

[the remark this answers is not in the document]
adlerjohn: There's two alternatives:
1. Such a tx is invalid, so it won't be included in a valid block.
1. The tx sender gets their fee back.

I prefer the former because it involves less work. What do you think?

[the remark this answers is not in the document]
Liamsi: Ok, before finalizing the block going through the Txs and checking if their corresponding message was also received and reject those whose message wasn't, seems much easier. So I would also prefer that (1.). My only concern is that one could effectively DoS the system by sending Tx without a message attached (as the mempool fills up with invalid Tx and there is no or very little cost in doing so).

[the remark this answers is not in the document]
adlerjohn: It should be easy to mitigate this DoS vector: just evict transactions that pay for a message unless they're accompanied by a message immediately. Or rather, don't even accept txs that pay for a message into the mempool unless the message is also provided.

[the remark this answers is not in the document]
Liamsi: Oh, it sounded like this were actually two steps. If it is one step and the message is sent along with the Tx (and not only a reference to the message as my original question was implying), we should be fine.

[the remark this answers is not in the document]
Liamsi: And this is good as this exactly how Cosmos SDK transactions work; they are a collection of messages: https://docs.cosmos.network/master/intro/sdk-design.html

Issue #8, opened 2020-03-23
Specify cryptographic primitives


Liamsi:
LazyLedger, as all blockchains uses numerous cryptographic primitives (most notably, hash functions and ecdsa or other cryptographic signature algorithms, e.g. BLS). 
We need to choose among these and document the choices made in the spec (and ideally document why these decisions where made somewhere else). 

Note: While these decisions are critical for a final spec, decisions about well established building blocks can happen at a later point in time in the specification process. We can and should start the implementation with the choices made in tendermint anyways to rapidly prototype.

adlerjohn:
Reopening since we need to more completely define the signature scheme, and how it interacts with consensus messages (e.g. votes).

PR #10, opened 2020-03-27
Spec out crypto primitives

adlerjohn:
Related issue: #8.

- [x] Public-key crypto encoding
- [ ] Vote serialization (https://github.com/LazyLedger/lazyledger-specs/pull/2#discussion_r377012152)
- [ ] Merkle trees (need a generic annotated Merkle tree)

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -206,6 +206,8 @@ Consensus-critical data is authenticated using [ECDSA](https://www.secg.org/sec1
 
 [Public key](#publickey)s are encoded in uncompressed form, as the concatenation of the `x` and `y` values. No prefix is needed to distinguish between encoding schemes as this is the only encoding supported.
 
+Deterministic signatures ([RFC-6979](https://tools.ietf.org/rfc/rfc6979.txt)) should be used when signing, but this is not enforced at the protocol level as it cannot be.
+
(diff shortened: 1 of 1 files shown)


specs/data_structures.md
```diff
@@ -195,9 +202,13 @@ Unless otherwise indicated explicitly, objects are first [serialized](#serializa
 
 # Public-Key Cryptography
 
-Consensus-critical data is authenticated using ECDSA, with the curve [secp256k1](https://en.bitcoin.it/wiki/Secp256k1). A highly-optimized library is available in C (https://github.com/bitcoin-core/secp256k1).
+Consensus-critical data is authenticated using [ECDSA](https://www.secg.org/sec1-v2.pdf), with the curve [secp256k1](https://en.bitcoin.it/wiki/Secp256k1). A highly-optimized library is available in C (https://github.com/bitcoin-core/secp256k1).
+
+[Public key](#publickey)s are encoded in uncompressed form, as the concatenation of the `x` and `y` values. No prefix is needed as this is the only encoding supported.
+
+[Signature](#signature)s are encoded as the `r` and `s` values of the signature, with one twist. Only low-`s` values in signatures are valid (i.e. `s < secp256k1.n`); `s` can be replaced with `-s mod secp256k1.n` during the signing process if it is high. Given this, the first bit of `s` will always be `0`, and can be used to store the 1-bit `v` value.
```
Liamsi: If mentioning r,s,v, it might be useful add a sentence about v as well? As far as I remember we dropped `v` in the tendermint implementation (can't remember the details right now). 

Don't sigs in ethereum consist of 65 bytes? If the main reason of using `Secp256k1` is ethereum compatibility, we should stick to whatever ethereum uses for signature representation. So I think this section should either be self-contained and replicate some details of eip-2098 (also about v), or, only refer to existing eips/other specs.

For an implementation, we can simply use ethereums go-wrapper around the mentioned C library.

adlerjohn: Expanded on the meaning of `v` in 14982e2. Solidity (or rather, the EVM) supports bitwise operations, so it's trivial to extract `v` from `vs`. Saving on passing in an extra word, in addition to history size reduction, is well worth it imo.

PR #12, opened 2020-04-01
Refactor header roots

adlerjohn:
- [x] evidence root
- [x] validators/state roots


Liamsi:
Rendered:
https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-refactor_header_roots/specs/data_structures.md
and
https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-refactor_header_roots/specs/architecture.md

adlerjohn:
@Liamsi Refactored the header to remove validator roots (both), and added a `stateCommitment` field, which is computed as `hash(stateRoot, validatorRoot)`. Does this seem reasonable and in line with previous discussion? If yes, I'll update the data structures figure and flag this for review.

Liamsi:
Still need to go through the tendermint changes required for this (https://github.com/LazyLedger/tendermint/issues/3) but independent of that this seems reasonable. 


Question: Is `hash` in `hash(stateRoot, validatorRoot)` the same hash function as used for the sparse merkle tree that yields, e.g. `stateRoot`

adlerjohn:
> Is hash in hash(stateRoot, validatorRoot) the same hash function as used for the sparse merkle tree that yields, e.g. stateRoot

Yes, my understanding is that only a single hash function is used everywhere (keccak-256).

Liamsi:
Do we need to prefix the `stateRoot` and the `validatorRoot` with the trees inner node prefix?

adlerjohn:
I don't think we _need_ to prefix it with anything, since it's a fixed operation. But it wouldn't hurt to just prefix it, that way it might be easier to parse, as it can be interpreted as the parent node of two potentially-asymmetric subtrees.

specs/data_structures.md
```diff
@@ -44,27 +44,24 @@ Blocks are the top-level data structure of the LazyLedger blockchain.
...
+| `height`            | `uint64`                  | Block height. The genesis block is at height `1`.                                            |
+| `time`              | [Time](#time)             | Timestamp of this block.                                                                     |
+| `lastBlockID`       | [BlockID](#blockid)       | Previous block's ID.                                                                         |
+| `lastCommitRoot`    | [HashDigest](#hashdigest) | Previous block's Tendermint commit root.                                                     |
+| `consensusHash`     | [HashDigest](#hashdigest) | Consensus parameters for this block.                                                         |
+| `stateCommitment`   | [HashDigest](#hashdigest) | Commitment to state root and validator set root after this block's transactions are applied. |
+| `availableDataRoot` | [HashDigest](#hashdigest) | Root of [commitments to erasure-coded data](#availabledataheader).                           |
+| `proposerAddress`   | [Address](#address)       | Address of this block's proposer.                                                            |
```
Liamsi, line 64: image

`lastBlockId` in the markdown text looks different? 🤔 

https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-refactor_header_roots/specs/data_structures.md#blockid

adlerjohn: Oh yeah, that arrow should be from the `availableDataRoot`

adlerjohn: Fixed in 7c85201

Liamsi:
From https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-refactor_header_roots/specs/architecture.md

> Transactions are signed and must be processed by clients to determine the validator set, while messages are un-signed data blobs that will usually represent an app's block data.

Didn't we settle on including messages inside of Tx?

adlerjohn:
> Didn't we settle on including messages inside of Tx?

I don't think so? Txs that pay fees for a message to be included (not all txs do this, some just e.g. change the validator set) shouldn't include the binary data as part of the tx itself, otherwise some client profiles would have to fully download messages even if they don't care about them. For example, if I only care about the LazyLedger validator set, I can just ignore the entire message namespace. Txs should only include a commitment to a message.

Liamsi:
I misunderstood the outcome of this discussion then: https://github.com/LazyLedger/lazyledger-specs/pull/2#discussion_r399303321

I thought it might be simple if we have a single Tx type `type Tx {msg: Msg}` and that model could also be used for validator set changes and everything. 

This is kinda orthogonal to this PR. We can continue the discussion in a separate issue. If I want to include a Msg, what do I do / what happens (and if sending Tx{commitToMsg} and Message,  is two operations couldn't I spam the network with unsigned messages that do not have a corresponding Tx)?

adlerjohn:
Transactions can certainly be sent around the network as `Tx {amount, recipient, witness[, msgBytes]}`, then encoded in the block as `Tx{amount, recipient, witness[, msgID]}` and `Msg{msgBytes}` (where `msgID = hash(msgbytes)`, and can be computed before relaying the transaction).

The point of splitting up txs and msgs _inside_ the block is that there can be users that want to ensure the integrity of the LazyLedger coin and make sure that the message data is available, but they don't care what the message data actually is.

Liamsi:
Makes perfect sense _inside_ the block! I was looking from an encoded on / decoded from the wire perspective I guess.

PR #16, opened 2020-04-12
Spec out Sparse Merkle Tree

adlerjohn:
- [x] Generic annotated Merkle tree
- [x] Sparse Merkle Tree

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -265,7 +265,7 @@ If a compact Merkle root is needed, the root level (which consists of root field
 
 ### Verifying Annotated Merkle Proofs
 
-
+In addition to the root, leaf, index, and sibling values of a Merkle proof for a plain [binary Merkle tree](#binary-merkle-tree), Merkle proofs for annotated Mekle trees have the sibling field values. Proofs are verified by using the appropriate methods to compute field values.
 
 ## Namespace Merkle Tree
 
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -247,9 +247,9 @@ Annotated Merkle trees have extra fields and methods to compute values for those
 
 For leaf node of leaf message `m`, its value `v` and fields `f_1, ..., f_n` are:
 ```C++
-f_1 = m_1_l(serialize(m))
+f_1 = m_1_l(m)
 ...
-f_n = m_n_l(serialize(m))
+f_n = m_n_l(m)
 v = h(serialize(m))
 ```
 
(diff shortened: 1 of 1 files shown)


Liamsi:
Rendered version: https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-smt_initial_spec/specs/data_structures.md#merkle-trees

specs/data_structures.md
```diff
@@ -224,49 +225,91 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
...
+For internal node at height `height` with children `l` and `r`, its value `v` and fields `f_1, ..., f_n` are:
+```C++
+f_1 = m_1_i(height, l.f_1, r.f_1)
+...
+f_n = m_n_i(height, l.f_n, r.f_n)
+v = h(l.f_1, ..., l.f_n, l.v, r.f_1, ..., r.f_n, r.v)
+```
+
+If a compact Merkle root is needed, the root level (which consists of root fields and a root value) can be hashed once.
+
```
Liamsi, line 265: Suggestion: Maybe add a sentence that (maybe how) annotating could also be used to differentiate between inner and leaf hashes?

specs/data_structures.md
```diff
@@ -224,49 +225,91 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
 
 ## Binary Merkle Tree
 
-Binary Merkle trees are constructed in the usual fashion, with leaves being hashed once to get leaf nodes and internal nodes being the hash of the concatenation of their children. Note that when hashing leaves , `0x00` is prepended, and when hashing internal nodes, `0x01` is prepended. This avoids a second-preimage attack [where internal nodes are presented as leaves](https://en.wikipedia.org/wiki/Merkle_tree#Second_preimage_attack).
+Binary Merkle trees are constructed in the usual fashion, with leaves being hashed once to get leaf node values and internal node values being the hash of the concatenation of their children.
```
Liamsi, line 228: While there might be an argument against differentiating internal and leaf nodes with a prefix in the sparse Merkle tree case (although all major SMT implementations still do that), this section speak about general binary trees. If you do not want to explicitly mention porefixing here, maybe say that leaf nodes are hashed with a hash function hash_leaf and inner nodes are hashed with another one hash_inner. In the case of no prefixes both are the same.

specs/data_structures.md
```diff
@@ -224,49 +225,91 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
...
 
-For leaf node of message `m`:
+For leaf node of leaf message `m`, its value `v` is:
 ```C++
-v = h(0x00, serialize(m))
+v = h(serialize(m))
 ```
 
-An exceptions is made, in the case of empty leaf nodes: the value of an empty leaf node is 32-byte zero, i.e. `0x0000000000000000000000000000000000000000000000000000000000000000`. This is used rather than duplicating the last node if there are an odd number of nodes in order to avoid [CVE-2012-2459](https://nvd.nist.gov/vuln/detail/CVE-2012-2459). Implicitly, trees are padded with empty nodes up to the next larger power of 2.
+An exception is made, in the case of empty leaves: the value of a leaf node with an empty leaf is 32-byte zero, i.e. `0x0000000000000000000000000000000000000000000000000000000000000000`. This is used rather than duplicating the last node if there are an odd number of nodes in order to avoid [CVE-2012-2459](https://nvd.nist.gov/vuln/detail/CVE-2012-2459). Implicitly, trees are padded with empty leaves up to the smallest enclosing power of 2.
```
Liamsi, line 236: >Implicitly, trees are padded with empty leaves up to the smallest enclosing power of 2.

The reason for this isn't CVE-2012-2459, right? Maybe add a sentence why the tree is padded with empty leaves?

adlerjohn: Yeah, good point. It's more to avoid the complexity of the Bitcoin design of duplicating odd nodes, which resulted in things like that CVE. Changing.

PR #21, opened 2020-04-20
Add erasure coding and available data to spec

adlerjohn:
Add the following:
- [x] Refactor available data section
- [x] Share format
- [x] Arrangement of available data in matrix
- [x] 2D matrix layout
- [x] How to compute commitments to available data in block header
- [x] Actual erasure coding scheme spec

Rendered:
* https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-erasure_coding_rsmt2d/specs/data_structures.md
* https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-erasure_coding_rsmt2d/specs/consensus.md


specs/data_structures.md
```diff
@@ -322,29 +285,111 @@ A proof into an SMT is structured as:
...
- | -------------- | ------------------------------- | --------------------- |
- | `transactions` | [Transaction](#transaction)`[]` | List of transactions. |
+### WrappedTransaction
 
-TODO define a Transaction format
+| name           | type          | description                                                                                     |
+| -------------- | ------------- | ----------------------------------------------------------------------------------------------- |
+| `index`        | `uint64`      | Index of this transaction in the list of wrapped transactions. This is needed for fraud proofs. |
+| `transaction`  | `Transaction` | Actual transaction.                                                                             |
+| `messageStart` | `uint64`      | _Optional_. Starting share of message this transaction pays for.                                |
```
Liamsi, line 308: Is that the share index?

adlerjohn: It should be the share index. In fact, with the "tiling" scheme, messages always start at the start of a share.

specs/figures/share.dot
```diff
@@ -0,0 +1,21 @@
...
+            <td width="79" align="left" border="0" cellpadding="0">1</td>
+            <td width="70" align="left" border="0" cellpadding="0">80</td>
+            <td width="106" align="left" border="0" cellpadding="0">150</td>
+            <td align="left" border="0" cellpadding="0">256</td>
+        </tr>
+        <tr>
+            <td width="1" cellpadding="4">*</td>
+            <td width="79" cellpadding="4">end of tx2</td>
+            <td width="70" cellpadding="4">tx3</td>
+            <td width="106" cellpadding="4">zero-padding</td>
```
musalbas, line 17: What padding scheme are we using?

specs/data_structures.md
```diff
@@ -320,31 +286,131 @@ A proof into an SMT is structured as:
 
 # Erasure Coding
 
+In order to enable trust-minimized light clients (i.e. light clients that do not rely on an honest majority of validating state assumption), it is critical that light clients can determine whether the data in each block is _available_ or not, without downloading the whole block itself. The technique used here was formally described in the paper [Fraud and Data Availability Proofs: Maximising Light Client Security and Scaling Blockchains with Dishonest Majorities](https://arxiv.org/abs/1809.09044).
```
musalbas, line 289: Technically in LazyLedger everyone uses data availability proofs, as there is no concept of a "light client" for the main chain.

specs/figures/share.dot
adlerjohn: Ah, I forgot to write that. It should be "pad with `0`s to fill the rest of the share."

specs/data_structures.md
adlerjohn: Hmm I guess in this case a "full node" would be one that fully validates the staking and fee part (basically, whatever's in the "reserved' namespace). We should probably write down specifically what different classes of nodes are and specifically what they are supposed to download, and what their assumptions and attack vectors are.

specs/figures/share.dot
adlerjohn: Fixed in 09522ba

specs/data_structures.md
```diff
@@ -320,31 +286,133 @@ A proof into an SMT is structured as:
...
+
+The remainder of the subsections below specify the [2D Reed-Solomon erasure coding scheme](#2d-reed-solomon-encoding-scheme) used, along with the format of [shares](#share) and how [available data](#available-data) is arranged into shares.
+
+## Reed-Solomon Erasure Coding
+
+TODO specify specifically
+
+## 2D Reed-Solomon Encoding Scheme
+
+![fig: RS2D encoding.](figures/rs2d.svg)
```
Liamsi, line 299: That file seems missing or has another name.

adlerjohn: Yeah, it's one I'm working on and forgot to remove this line before pushing the commit!

specs/data_structures.md
adlerjohn: Fixed in 71afd1f.

specs/data_structures.md
```diff
@@ -320,31 +286,172 @@ A proof into an SMT is structured as:
...
 # Erasure Coding
 
+In order to enable trust-minimized light clients (i.e. light clients that do not rely on an honest majority of validating state assumption), it is critical that light clients can determine whether the data in each block is _available_ or not, without downloading the whole block itself. The technique used here was formally described in the paper [Fraud and Data Availability Proofs: Maximising Light Client Security and Scaling Blockchains with Dishonest Majorities](https://arxiv.org/abs/1809.09044).
+
+The remainder of the subsections below specify the [2D Reed-Solomon erasure coding scheme](#2d-reed-solomon-encoding-scheme) used, along with the format of [shares](#share) and how [available data](#available-data) is arranged into shares.
+
+## Reed-Solomon Erasure Coding
+
+TODO specify specifically
+
```
adlerjohn, line 296: This part of the erasure coding spec can be left for later as it's a design detail.

adlerjohn: NVM, added in 5c61e04.

adlerjohn:
Okay, take a look @Liamsi, should be good now.

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+In order to enable trust-minimized light clients (i.e. light clients that do not rely on an honest majority of validating state assumption), it is critical that light clients can determine whether the data in each block is _available_ or not, without downloading the whole block itself. The technique used here was formally described in the paper [Fraud and Data Availability Proofs: Maximising Light Client Security and Scaling Blockchains with Dishonest Majorities](https://arxiv.org/abs/1809.09044).
+
+The remainder of the subsections below specify the [2D Reed-Solomon erasure coding scheme](#2d-reed-solomon-encoding-scheme) used, along with the format of [shares](#share) and how [available data](#available-data) is arranged into shares.
+
+## Reed-Solomon Erasure Coding
+
+Reed-Solomon erasure coding is used as the underlying coding scheme. The parameters are:
+- 16-bit Galois field
+- 1:1 original:parity pieces
+- `AVAILABLE_DATA_MATRIX_HALFSIZE` original pieces (i.e `AVAILABLE_DATA_MATRIX_HALFSIZE*2` total pieces)
```
Liamsi, line 298: Oh, I see. `AVAILABLE_DATA_MATRIX_HALFSIZE` is the number of original pieces (called `k` below). Total pieces refers to original pieces and parity pieces I assume? This is a complicated way to express above's relation again (1:1 original:parity pieces).

It is confusing to read sth about a matrix here already as this introduces 1d-RS?

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+- `SHARE_SIZE` bytes per piece
+
+[Leopard-RS](https://github.com/catid/leopard) is a C library that implements the above scheme with quasilinear runtime.
+
+## 2D Reed-Solomon Encoding Scheme
+
+The 2-dimensional data layout is described in this section. Data is arranged in `k * k` matrix. The roots of [NMTs](#namespace-merkle-tree) for each row and column across four quadrants of data, `Q0` to `Q3` must be computed. In other words, `2k` row roots and `2k` column roots must be computed. The row and column roots are stored in the `availableDataCommitments` of the [AvailableDataHeader](#availabledataheader).
```
Liamsi, line 305: Why not talk about `AVAILABLE_DATA_MATRIX_HALFSIZE` in relation to `k` here instead of above?

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+
+![fig: RS2D encoding: data quadrants.](figures/rs2d_quadrants.svg)
+
+The data of `Q0` is the original data, and the remaining quadrants are parity data. Where `A -> B` indicates that `B` is computed using [erasure coding](#reed-solomon-erasure-coding) from `A`:
+- `Q0 -> Q1` for each row in `Q0` and `Q1`
+- `Q0 -> Q2` for each column in `Q0` and `Q2`
+- `Q2 -> Q3` for each row in `Q2` and `Q3`
```
Liamsi, line 312: 👍

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+
+## Share
+
+A share is a fixed-size data chunk that will be erasure-coded and committed to in [Namespace Merkle trees](#namespace-merkle-tree).
+
+| name      | type               | description     |
+| --------- | ------------------ | --------------- |
+| `rawData` | `byte[SHARE_SIZE]` | Raw share data. |
+
+An example layout of the share's internal bytes is shown below. For non-parity shares _with a reserved namespace_, the first `SHARE_RESERVED_BYTES` bytes (`*` in the figure) is the starting byte of the first request in the share as an unsigned integer, or `0` if there is none. In this example, the first byte would be `80` (or `0x50` in hex). For shares _with a non-reserved namespace_ (and parity shares), the first `SHARE_RESERVED_BYTES` bytes have no special meaning and are simply used to store data like all the other bytes in the share.
```
Liamsi, line 336: > is the starting byte of the first request in the share as an unsigned integer,

Now I see what you meant with starting byte. I find the word request in this context confusing. Wouldn't it be simpler if all `tx` wher length prefixed? Then you'd know when tx2 will end and tx3 will start.

specs/consensus.md
```diff
@@ -25,12 +27,29 @@ Consensus Rules
...
-| ---------------------- | -------- | ------- | ------ | ------------------------------------------------- |
-| `NAMESPACE_ID_BYTES`   | `uint64` | `32`    | `byte` | Size of namespace ID, in bytes.                   |
-| `SHARE_SIZE`           | `uint64` | `256`   | `byte` | Size of transaction and message shares, in bytes. |
-| `SHARE_RESERVED_BYTES` | `uint64` | `1`     | `byte` | Bytes reserved at the beginning of each share.    |
-| `GENESIS_COIN_COUNT`   | `uint64` | `10**8` | `4u`   | `(= 100000000)` Number of coins at genesis.       |
+| name                             | type     | value   | unit    | description                                                                                                          |
+| -------------------------------- | -------- | ------- | ------- | -------------------------------------------------------------------------------------------------------------------- |
+| `NAMESPACE_ID_BYTES`             | `uint64` | `64`    | `byte`  | Size of namespace ID, in bytes.                                                                                      |
```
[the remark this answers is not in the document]
adlerjohn: Me thinking that it said "bits" instead of "bytes" and wanting to raise it from 32 bits to a more reasonable 64 bits. 32 bytes (256 bits) was perfectly fine. 😞

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
 
+In order to enable trust-minimized light clients (i.e. light clients that do not rely on an honest majority of validating state assumption), it is critical that light clients can determine whether the data in each block is _available_ or not, without downloading the whole block itself. The technique used here was formally described in the paper [Fraud and Data Availability Proofs: Maximising Light Client Security and Scaling Blockchains with Dishonest Majorities](https://arxiv.org/abs/1809.09044).
+
+The remainder of the subsections below specify the [2D Reed-Solomon erasure coding scheme](#2d-reed-solomon-encoding-scheme) used, along with the format of [shares](#share) and how [available data](#available-data) is arranged into shares.
+
+## Reed-Solomon Erasure Coding
+
+Reed-Solomon erasure coding is used as the underlying coding scheme. The parameters are:
+- 16-bit Galois field
+- 1:1 original:parity pieces
```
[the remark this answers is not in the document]
adlerjohn: Technically you can have an erasure coding where you need e.g 80% of the total shares to recover. This parameter is equivalent to saying "50%."

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
- | -------------- | ------------------------------- | --------------------- |
- | `transactions` | [Transaction](#transaction)`[]` | List of transactions. |
+### WrappedTransaction
 
-TODO define a Transaction format
+| name           | type          | description                                                                                       |
+| -------------- | ------------- | ------------------------------------------------------------------------------------------------- |
+| `index`        | `uint64`      | Index of this transaction in the list of wrapped transactions. This is needed for fraud proofs.   |
+| `transaction`  | `Transaction` | Actual transaction.                                                                               |
+| `messageStart` | `uint64`      | _Optional_. Share index (in row-major order) of first share of message this transaction pays for. |
```
Liamsi, line 376: This needs a better name. Maybe `messageIndex` or `msgStartIndex`?

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
 
-Wrapper for transaction data, which is a simple list of [Transaction](#transaction)s.
+| name                  | type                                          | description                   |
+| --------------------- | --------------------------------------------- | ----------------------------- |
+| `wrappedTransactions` | [WrappedTransaction](#wrappedtransaction)`[]` | List of wrapped transactions. |
 
- | name           | type                            | description           |
- | -------------- | ------------------------------- | --------------------- |
- | `transactions` | [Transaction](#transaction)`[]` | List of transactions. |
+### WrappedTransaction
```
Liamsi, line 370: I think it would be good to motivate the `wrapped` types in a sentence before.
Sth like: In order to be able to lookup the corresponding messages a Tx pays for in the "data matrix" we introduce the notion of wrapped Tx.

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
 
-TODO define a Transaction format
+| name           | type          | description                                                                                       |
+| -------------- | ------------- | ------------------------------------------------------------------------------------------------- |
+| `index`        | `uint64`      | Index of this transaction in the list of wrapped transactions. This is needed for fraud proofs.   |
```
Liamsi, line 374: It's not clear to me how exactly this is needed for fraud proofs.

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+| TODO                    |                           |                                                                                                                                                                                                                                                                       |
+| `messageSize`           | `uint64`                  | Size of message this transaction pays a fee for, in `byte`s. If this transaction does not pay a fee for a message, must be `0`.                                                                                                                                       |
+| `messageShareRootsRoot` | [HashDigest](#hashdigest) | Merkle root of message share roots of an optional message that this transaction pays a fee to be included in the current block. Messages are split into shares and committed to here. Large messages can span across rows, which requires more roots to the provided. |
```
Liamsi, line 384: Why do these roots need to be included in the Tx?

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+
+| name                            | type                                                              | description                               |
+| ------------------------------- | ----------------------------------------------------------------- | ----------------------------------------- |
+| `wrappedIntermediateStateRoots` | [WrappedIntermediateStateRoot](#wrappedintermediatestateroot)`[]` | List of wrapped intermediate state roots. |
+
+### WrappedIntermediateStateRoot
+
+| name                    | type         | description                                                                                                     |
+| ----------------------- | ------------ | --------------------------------------------------------------------------------------------------------------- |
+| `index`                 | `uint64`     | Index of this intermediate state root in the list of intermediate state roots. This is needed for fraud proofs. |
```
Liamsi, line 397: This probably will become clearer later when fraud proofs are described in detail.

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+| ----------- | ------------------------- | ---------------------------------------------- |
+| `evidences` | [Evidence](#evidence)`[]` | List of evidence used for slashing conditions. |
+
+### Evidence
+
+| name     | type                    | description |
+| -------- | ----------------------- | ----------- |
+| `pubKey` | [PublicKey](#publickey) |             |
+| `voteA`  | [Vote](#vote)           |             |
+| `voteB`  | [Vote](#vote)           |             |
```
Liamsi, line 420: If this is the only kind of evidence we will ever use (currently not 100% clear to me), we can be more specific in the naming everywhere (e.g. `DoubleSigningEvidence`)

specs/data_structures.md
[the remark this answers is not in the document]
Liamsi: Ok, I see. This probably has a name in the erasure coding world. Best to use that word instead. So I guess the param is 2 (meaning you can loose half the data / need half the data to recover).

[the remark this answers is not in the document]
Liamsi: What you are describing seems to be called _erasure efficiency_.
> The fraction k’/k, where k’ denotes the number of symbols required for recovery, is called reception efficiency.
So in the general sense the parameter would be the the desired erasure efficiency (here 1/2).

In the context of reed solomon this is simply defined by the number of check symbols (or parity pieces). So the parameter is the number of parity symbols:

> By adding t check symbols to the data, a Reed–Solomon code can detect (but not correct) any combination of up to and including t erroneous symbols, OR locate and correct up to and including ⌊t/2⌋ erroneous symbols at unknown locations. As an erasure code, it can correct up to and including t erasures at locations that are known and provided to the algorithm, or it can detect and correct combinations of errors and erasures.

If we fix one half, the only actual param is the number of original data symbols (and maybe the symbol size).

specs/data_structures.md
Liamsi: Thinking about this further, Evidence should probably be an enumerated type (currently only containing DoubleSigningEvidence as a member). Tendermint currently uses amino prefixing to encode the concrete instantiations of the Evidence interface but I think this will change to proto's `oneof` very soon.

adlerjohn: The particularities of evidence are outside the scope of this PR, but yes I agree that we'll need some evidence types (double signing is one, but we should also have one for proposing an invalid block---hence, fraud proofs).

specs/consensus.md
```diff
@@ -25,12 +27,29 @@ Consensus Rules
...
+| `NAMESPACE_ID_RESERVED_BYTES`    | `uint64` | `2`     | `byte`  | Size of reserved namespace ID range, in bytes.                                                                       |
+| `SHARE_SIZE`                     | `uint64` | `256`   | `byte`  | Size of transaction and message shares, in bytes.                                                                    |
+| `SHARE_RESERVED_BYTES`           | `uint64` | `1`     | `byte`  | Bytes reserved at the beginning of each share. Must be sufficient to represent `SHARE_SIZE`.                         |
+| `AVAILABLE_DATA_MATRIX_HALFSIZE` | `uint64` |         | `share` | Size of the data availability matrix (analogous to blocksize). Half the number of shares on each side of the matrix. |
```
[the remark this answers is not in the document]
adlerjohn: Yes, this is the `k`. I agree the description is really unclear, fixing.

specs/data_structures.md
adlerjohn: I used "request" here in the context of the [Tendermint paper](https://arxiv.org/abs/1807.04938), which is equivalent to "transactions" for a distributed ledger. Probably not the best word. Any ideas for something that can describe "any entry in the ledger"?

> Wouldn't it be simpler if all tx were length prefixed? Then you'd know when tx2 will end and tx3 will start.
You mean just if every request in a share had a length prefix? I don't know if that would be necessarily be simpler, since the number of "length bytes" would be variable instead of fixed. With this scheme you can just take 255-byte slices of the serialized requests.

specs/data_structures.md
```diff
@@ -320,31 +286,178 @@ A proof into an SMT is structured as:
...
+1. Place the first share of the message at the next unused location in the matrix whose column in aligned with the number of padded shares (i.e. if there are four padded shares, then only every fourth location can be used to start) **unless** there are insufficient unused locations in the row. Then place the remaining shares in the following locations.
+1. If there are insufficient unused locations in the row, place the first share of the message at the first column of the next row. Then place the remaining shares in the following locations.
+
+In the example below, two messages (of padded sizes 2 and 1) are placed following the aforementioned rules.
+
+![fig: Original data: messages.](figures/rs2d_originaldata_message.svg)
+
+By construction, this gives a useful property: transactions [can commit to a commitment to a list of roots](#transaction) that are each guaranteed (assuming the block is valid) to be subtree roots in one or more of the row NMTs.
```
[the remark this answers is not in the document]
adlerjohn: If `k` is a power of 2, then using the two rules guarantees (unless I'm mistaken) that the transaction sender knows enough about the layout of the message in a valid block to know how it will be split up across rows (if it is) and thus how to commit to it.

As an example, suppose `k = 8` and the number of shares of the message is `16`. The transaction sender knows it'll take up two rows (they don't know which two exactly, just some two). Since they commit to row subtree roots, they can just commit to the roots of those two rows, then the root of those roots.

To verify without downloading all the message data, we know that a message of size `16` takes up two rows. So we can ask for both subtree roots that are behind the hash that's in the transaction, and check that they're included at the rows indicated by `messageStart` that the block producer sets.

specs/data_structures.md
adlerjohn: Those two fields are optional, and only included if a transaction pays for a message to be included. If we want fees to be consensus-critical (a prerequisite for burning some coins for each unit work a block requires), then we need the transaction sender to commit to this information.

specs/data_structures.md
Liamsi: > You mean just if every request in a share had a length prefix?

Yes (basically len(entry )|| entry, for each entry.

>  I don't know if that would be necessarily be simpler, since the number of "length bytes" would be variable instead of fixed. With this scheme you can just take 255-byte slices of the serialized requests.

Hmm, good point. Let's keep it as is. I'll implement this and if there is any insights from that we can still iterate in a separate PR.

specs/consensus.md
```diff
@@ -27,14 +27,14 @@ Consensus Rules
...
+| name                                  | type     | value   | unit    | description                                                                                  |
+| ------------------------------------- | -------- | ------- | ------- | -------------------------------------------------------------------------------------------- |
+| `NAMESPACE_ID_BYTES`                  | `uint64` | `32`    | `byte`  | Size of namespace ID, in bytes.                                                              |
+| `NAMESPACE_ID_RESERVED_BYTES`         | `uint64` | `2`     | `byte`  | Size of reserved namespace ID range, in bytes.                                               |
+| `SHARE_SIZE`                          | `uint64` | `256`   | `byte`  | Size of transaction and message shares, in bytes.                                            |
+| `SHARE_RESERVED_BYTES`                | `uint64` | `1`     | `byte`  | Bytes reserved at the beginning of each share. Must be sufficient to represent `SHARE_SIZE`. |
+| `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE` | `uint64` |         | `share` | Number of rows/columns of the original data shares in square layout.                         |
```
Liamsi: 👍

specs/data_structures.md
```diff
@@ -292,10 +292,12 @@ The remainder of the subsections below specify the [2D Reed-Solomon erasure codi
...
 ## Reed-Solomon Erasure Coding
 
+Note that while data is laid out in a two-dimensional square, rows and columns are erasure coded using a standard one-dimensional encoding.
+
 Reed-Solomon erasure coding is used as the underlying coding scheme. The parameters are:
 - 16-bit Galois field
-- 1:1 original:parity pieces
-- `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE` original pieces (i.e `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE*2` total pieces)
+- `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE` original pieces
+- `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE` parity pieces (i.e `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE*2` total pieces), for an erasure efficiency of 50%
```
Liamsi: Maybe still clarify if this means 50% of total data (orig || parity), or, separately (50% of orig /  50% of parity respectively).

adlerjohn: Clarified in cd3b85d.

PR #25, opened 2020-05-10
Add protobuf defs for data structures

adlerjohn:
Add compilable protobuf3 files for data structures (anything that will need to go over the wire).


proto/types.proto
```diff
@@ -0,0 +1,140 @@
...
+  bytes raw_data = 1;
+}
+
+message Address {
+  bytes raw_data = 1;
+}
+
+message CommitSig {
+  enum BlockIDFlag {
+    BLOCK_ID_FLAG_UNSPECIFIED = 0;
```
Liamsi, line 58: This is probably just copied and adapted from the tendermint protos but I'm wondering about the  the rationale behind this extra value. Shouldn't every value besides 1,2,3 unspecified/invalid?

I'll ping @marbar3778 about it.

proto/types.proto
```diff
@@ -0,0 +1,140 @@
...
+
+message IntermediateStateRoot {
+  HashDigest root = 1;
+}
+
+message EvidenceData {
+  repeated Evidence evidences = 1;
+}
+
+message Evidence {
```
Liamsi, line 106: Unrelated to this PR but note that with a more complete light client impl and ibc around the corner, tendermint now has these Evidence implementations:
```
Implementations of Evidence  (8 usages found)
    types  (8 usages found)
        evidence.go  (7 usages found)
            119 type DuplicateVoteEvidence struct {
            304 type ConflictingHeadersEvidence struct {
            557 type PhantomValidatorEvidence struct {
            660 type LunaticValidatorEvidence struct {
            797 type PotentialAmnesiaEvidence struct {
            921 type MockRandomEvidence struct {
            943 type MockEvidence struct {
        vote.go  (1 usage found)
            29 type ErrVoteConflictingVotes struct {
```

Weirdly the proto files list `Evidence` as a oneof of two types 🤔:
```proto
message Evidence {
  oneof sum {
    DuplicateVoteEvidence duplicate_vote_evidence = 1;
    MockEvidence          mock_evidence           = 2;
  }
}
```

proto/types.proto
```diff
@@ -0,0 +1,140 @@
...
+message WrappedTransaction {
+  uint64 index = 1;
+  Transaction transaction = 2;
+  uint64 message_start_index = 3;
+}
+
+message Transaction {
+  uint64 message_size = 1;
+  HashDigest message_share_roots_root = 2;
+}
```
Liamsi, line 87: As far as I remembered, not all transactions need to refer to messages. But if they don't where does their data live?

proto/types.proto
Liamsi: Ah NVM, this is just a proto detail:
> - There must be a zero value, so that we can use 0 as a numeric default value.
> - The zero value needs to be the first element, for compatibility with the proto2 semantics where the first enum value is always the default.

https://developers.google.com/protocol-buffers/docs/proto3#enum

proto/types.proto
marbar3778: This has been updated. If you look at the proto branch now it will have more evidence types

Liamsi: Cool, can't wait for this to get merged ;-)

Liamsi: @adlerjohn in case you are wondering which branch Marko is referring to: https://github.com/tendermint/tendermint/blob/proto-breakage/proto/types/evidence.proto

proto/types.proto
adlerjohn: Yep, that was it.

proto/types.proto
adlerjohn: Transactions and evidence will be updated when those are fleshed out later.

PR #28, opened 2020-05-14
Clear up the namespace ID bound constant

adlerjohn:
Instead of a byte size (which doesn't really indicated anything), explicitly define the bound on reserved namespace IDs as a max value.


specs/consensus.md
```diff
@@ -32,7 +32,7 @@ Consensus Rules
 | `VERSION`                             | `uint64` | `1`     |         | Version of the LazyLedger chain. Breaking changes (hard forks) must update this parameter.                                                                  |
 | `CHAIN_ID`                            | `uint64` | `1`     |         | Chain ID. Each chain assigns itself a (unique) ID.                                                                                                          |
 | `NAMESPACE_ID_BYTES`                  | `uint64` | `32`    | `byte`  | Size of namespace ID, in bytes.                                                                                                                             |
-| `NAMESPACE_ID_RESERVED_BYTES`         | `uint64` | `2`     | `byte`  | Size of reserved namespace ID range, in bytes.                                                                                                              |
+| `NAMESPACE_ID_MAX_RESERVED`           | `uint64` | `65535` |         | Value of maximum reserved namespace ID (inclusive). 2 bytes worth of IDs.                                                                                   |
```
Liamsi, line 35: Wouldn't one byte be sufficiently large for reserved namespaces?

adlerjohn: Maybe? It doesn't really affect anything either way if we use hashes to assign namespace IDs. Just through it would be better to go liberal so we don't have to change it later, but the argument could be made to not be needlessly wasteful.

adlerjohn: If you feel anything stronger than pure apathy at the choice, then I can change it to 255 😂

adlerjohn: Reduced in 7d08046.

Liamsi: I didn't feel strong about it. I was just wondering.

PR #32, opened 2020-05-19
Add state representation

adlerjohn:
Define the state and the validator set.

Rendered:
- https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-state_representation/specs/data_structures.md
- https://github.com/LazyLedger/lazyledger-specs/blob/adlerjohn-state_representation/specs/consensus.md
- https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-state_representation/rationale/distributing_rewards.md


specs/data_structures.md
```diff
@@ -475,7 +477,28 @@ enum VoteType : uint8_t {
...
+
+Two [Sparse Merkle Trees](#sparse-merkle-tree) are maintained: one of [accounts](#account) and one of [validators](#validator). The state root is computed as the [hash](#hashdigest) of concatenation of the account tree root and the validator tree root.
+
+## Account
+
+| name                 | type                | description                                                                              |
+| -------------------- | ------------------- | ---------------------------------------------------------------------------------------- |
+| `balance`            | `uint64`            | Coin balance.                                                                            |
+| `isDelegating`       | `bool`              | Whether this account is delegating its stake or not.                                     |
+| `delegatedValidator` | [Address](#address) | _Optional._ The validator this is account is delegating to.                              |
```
Liamsi, line 490: Why can't accounts delegate to several validators?

adlerjohn: Because that would add complexity (you'd have to hold a variable number of validators and the number of coins each of them is delegated), and if someone wants to delegate their coins to two validators they can just...split up their coins into two accounts.

adlerjohn: This scheme (which isn't obvious from the data structures, but will be written in the consensus rules) is: "all the coins in an account are either delegated or not; if they're delegated they must be delegated to a single validator."

Liamsi: From a user perspective holding coins and a UX perspective, splitting up coins into separate accounts whenever you want to delegate to more than one validator also adds complexity but one the human (not on the software).

Note that in the few delegations based PoS systems  that launched, you can usually delegate to one or more validators. This is also true in Cosmos / the cosmos-sdk. Most users I know delegate to several validators to keep the network decentralized (and to hedge their risks).

adlerjohn: Hmm. I don't see why this can't be abstracted away at the application layer, rather than having to be pushed to the user layer. Moving a portion of your coins to a new address then delegating them (as well as the reverse) can be done just as seamlessly in a wallet---one button press. Users don't even have to know it's happening! It's basically UTXO management; most Bitcoin wallets don't present to lay users each individual UTXO, just their total balance.

Allowing an arbitrary number of delegations per account means fraud proofs are more complex and expensive.

That being said, given that we're using the accounts data model for now, one downside of essentially emulating UTXOs for delegating stake is that the nonce of empty accounts still needs to be kept around forever (which reminds me, a nonce field needs to be added).

adlerjohn:
Things that have been intentionally excluded:
1. Slashing. I suggest we use correlated slashing (i.e. increase the slash rate as more stake is penalized simultaneously), which might require reorganizing where this information is kept in state.
1. Multiple delegations or one-step re-delagations in-protocol (https://github.com/lazyledger/lazyledger-specs/pull/32#discussion_r428031498). Those should be able to be abstracted awake to the wallet/application layer, and we should avoid any protocol complexity that isn't strictly needed.
1. Complete state machine for validator and delegation objects. That will be specified in a later PR that fully details the consensus rules around that.

rationale/distributing_rewards.md
```diff
@@ -0,0 +1,54 @@
...
+In other words, the voting power contributed by the delegator multiplied by the _reward rate_, i.e. the rewards per unit of voting power. We note that if the total voting power of a validator remains constant forever, then the above equations hold and is approximation-free. However, changes to the total voting power need to be accounted for.
+
+Blocks between two changes to a validator's voting power (i.e. whenever a user delegates or undelegates stake) are a _period_. Every time a validator's voting power changes (i.e. a new period $f$ begins), an entry $Entry_f$ for this period is saved in state, which records _the reward rate up to the beginning of_ $f$. This is simply the sum of the reward rate up to the beginning of previous period $f-1$ and the reward rate of the period $f$ itself:
+
+$$
+Entry_f = \begin{cases}
+    0 & f = 0 \\
+    Entry_{f-1} + \frac{T_f}{n_f} & f > 0 \\
+\end{cases}
+$$
```
Liamsi, line 38: Is the plan to render this using some JS lib? This isn't nice to read when rendered as html currently because it is just a single line:
image
I rendered it here using https://www.quicklatex.com/
image

specs/consensus.md
```diff
@@ -66,6 +70,76 @@ Consensus Rules
 
 # Block Validity
 
+## State Transitions
+
+### Validators and Delegations
+
+A transaction `tx` that requests a new validator initializes the [Validator](data_structures.md#validator) field of that account as follows:
```
Liamsi, line 77:
```suggestion
A transaction `tx` that requests a new validator initializes the [Validator](data_structures.md#validator) fields of that account as follows:
```

specs/consensus.md
```diff
@@ -66,6 +70,76 @@ Consensus Rules
...
+| name              | value                    |
+| ----------------- | ------------------------ |
+| `status`          | `ValidatorStatus.Queued` |
```
Liamsi, line 80: This Queued status is necessary because we need to track this before the block is finalized for enabling fraud proofs?

specs/data_structures.md
```diff
@@ -475,7 +480,91 @@ enum VoteType : uint8_t {
 
 # State
 
-TODO validator set repr
+| name              | type                      | description                  |
+| ----------------- | ------------------------- | ---------------------------- |
+| `accountTrieRoot` | [HashDigest](#hashdigest) | Merkle root of account trie. |
```
Liamsi, line 485: Can't we simply call this accountRoot? And remove the reference to trie below too?

rationale/distributing_rewards.md
```diff
@@ -0,0 +1,54 @@
...
+
+This forms the primary motivation of the scheme discussed here: a mechanism for distributing rewards that is state-efficient and requires no iteration over state elements for any state transition.
+
+# Distribution Scheme
+
+The scheme presented here is an incarnation of Cosmos' [F1 fee distribution scheme](https://github.com/cosmos/cosmos-sdk/blob/master/docs/spec/_proposals/f1-fee-distribution/f1_fee_distr.pdf). F1 has the nice property of being approximation-free and, with proper implementation details, can be highly efficient with state usage and completely iteration-free in all cases.
```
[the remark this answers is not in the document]
adlerjohn: Echoing what I said on Slack: after more reading it turns out F1 can be used directly, so there are no design changes to it. Only implementation details are specified to improve performance.

rationale/distributing_rewards.md
adlerjohn: Right, that's the plan (e.g. https://github.com/mathjax/MathJax). Formulas like this will be limited to only be used in rationale docs, not specs.

specs/consensus.md
adlerjohn: Cosmos uses:
- `Unbonded -> Bonded -> Unbonding -> Unbonded`
Compared to:
- `Queued -> Bonded -> Unbonding -> Unbonded`

It's just a disambiguation of the first state.

specs/data_structures.md
adlerjohn: Fixed in e8cc2d3.

specs/consensus.md
Liamsi: Sounds reasonable (and `Queued` is clearer here). But this seems like an optional implementation detail.

specs/consensus.md
```diff
@@ -66,6 +70,76 @@ Consensus Rules
...
+| -------- | ------------------------ |
+| `status` | `ValidatorStatus.Bonded` |
+
+For validators that were bonded but are no longer (either by being outside the top validators or through a transaction that requests unbonding), they begin unbonding.
+| name              | value                       |
+| ----------------- | --------------------------- |
+| `status`          | `ValidatorStatus.Unbonding` |
+| `unbondingHeight` | `block.height + 1`          |
+
+Once an unbonding validator has waited at least `UNBONDING_DURATION` blocks, they can be unbonded, collecting their reward:
```
Liamsi, line 101: That sounds like they can only collect rewards when completely unbonded?

specs/consensus.md
```diff
@@ -66,6 +70,76 @@ Consensus Rules
...
+| `status`          | `DelegationStatus.Unbonding` |
+| `endEntry`        | `validator.latestEntry`      |
+| `unbondingHeight` | `block.height + 1`           |
+
+then updates the target validator's voting power:
+| name             | value                                      |
+| ---------------- | ------------------------------------------ |
+| `delegatedCount` | `old.delegatedCount - 1`                   |
+| `votingPower`    | `old.votingPower - delegation.votingPower` |
+
```
Liamsi, line 142: On the long run, we should create a state transitions diagrams.

specs/data_structures.md
musalbas: I feel that it would actually be more intuitive from a UX perspective to require users to split their coins into different accounts to delegate to different validators. It's the logical purpose of accounts - the same way you might have multiple savings accounts for different purposes.

It seems more complicated from a UX perspective to have users enter each validator they want to delegate to and the amount in some kind of table.

rationale/distributing_rewards.md
```diff
@@ -0,0 +1,54 @@
...
+
+Naively, when considering a single block, the reward that should be given to a delegator with stake $x$, who's delegating to a validator with total voting power $n$, whose reward in that block is $T$, is:
+
+$$
+\text{naive reward} = x \frac{T}{n}
+$$
+
+In other words, the voting power contributed by the delegator multiplied by the _reward rate_, i.e. the rewards per unit of voting power. We note that if the total voting power of a validator remains constant forever, then the above equations hold and is approximation-free. However, changes to the total voting power need to be accounted for.
+
+Blocks between two changes to a validator's voting power (i.e. whenever a user delegates or undelegates stake) are a _period_. Every time a validator's voting power changes (i.e. a new period $f$ begins), an entry $Entry_f$ for this period is saved in state, which records _the reward rate up to the beginning of_ $f$. This is simply the sum of the reward rate up to the beginning of previous period $f-1$ and the reward rate of the period $f$ itself:
```
musalbas, line 31: Why two changes? Shouldn't it be one change?

rationale/distributing_rewards.md
```diff
@@ -0,0 +1,54 @@
...
+
+$$
+\text{reward} = x (Entry_f - Entry_k)
+$$
+
+This raw reward can be scaled by additional factors, such as commissions or slashing penalties.
+
+## State-Efficient Implementation
+
+The F1 paper doesn't specify where entries are stored in state, but the understanding is that they are placed in independent state elements. This has the downside of requiring multiple Merkle branches to prove the inclusion of entries for e.g. fraud proofs. We can improve on this by leveraging a specific property of entries, namely that each entry is only used in exactly two cases:
```
musalbas, line 50: I think it's better to stick to formal language for consistency.

specs/consensus.md
```diff
@@ -66,6 +70,76 @@ Consensus Rules
...
+Once an unbonding validator has waited at least `UNBONDING_DURATION` blocks, they can be unbonded, collecting their reward:
+| name            | value                                 |
+| --------------- | ------------------------------------- |
+| `status`        | `ValidatorStatus.Unbonded`            |
+| `stakedBalance` | `0`                                   |
+| `votingPower`   | `old.votingPower - old.stakedBalance` |
+
+Every time a bonded validator's voting power changes (i.e. when a delegation is added or removed), or when a validator begins unbonding, the rewards per unit of voting power accumulated so far are calculated:
+| name             | value                                                    |
```
musalbas, line 109: What happens with these fields / where are they stored?

specs/data_structures.md
```diff
@@ -475,7 +480,91 @@ enum VoteType : uint8_t {
...
 
-TODO validator set repr
+| name            | type                      | description                  |
+| --------------- | ------------------------- | ---------------------------- |
+| `accountsRoot`  | [HashDigest](#hashdigest) | Merkle root of account tree. |
+| `numValidators` | `uint32`                  | Number of active validators. |
+
+The state of the LazyLedger chain contains only account balances and the validator set (which is extra metadata on top of the plain account balances).
+
+One unified [Sparse Merkle Trees](#sparse-merkle-tree) is maintained for both account account balances and validator metadata, the _accounts tree_. The final state root is computed as the [hash](#hashdigest) of the accounts tree root and number of active validators. The latter is necessary to ensure light nodes can determine the entire validator set from a single state root commitment.
```
musalbas, line 490: How are accounts and validator metadata keyed?

specs/data_structures.md
Liamsi: > It's the logical purpose of accounts - the same way you might have multiple savings accounts for different purposes.

But here you'd have multiple accounts for one purpose: namely delegate to validators. Having a different account per validator would be like opening a broker account per same type of investment.

adlerjohn:
Major changes since last review:
1. Changed tables to code snippets for consensus rules.
1. Use state subtrees to store accounts, active validator set, active validator count, and inactive validator set. The latter is used to make all delegation operations on inactive validators quite a bit cheaper (less hashing and guaranteed parallelization of hashing).

specs/consensus.md
```diff
@@ -57,6 +60,14 @@ Consensus Rules
 | `EVIDENCE_NAMESPACE_ID`                | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000003` |
 | `PARITY_SHARE_NAMESPACE_ID`            | `NamespaceID` | `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF` |
 
+## Reserved State Subtree IDs
+
+| name                         | type             | value  |
+| ---------------------------- | ---------------- | ------ |
+| `ACCOUNTS_SUBTREE_ID`        | `StateSubtreeID` | `0x01` |
+| `VALIDATORS_SUBTREE_ID`      | `StateSubtreeID` | `0x02` |
+| `VALIDATOR_COUNT_SUBTREE_ID` | `StateSubtreeID` | `0x03` |
```
Liamsi: Shouldn't this be part of the validator subtree?

specs/consensus.md
```diff
@@ -57,6 +60,14 @@ Consensus Rules
 | `EVIDENCE_NAMESPACE_ID`                | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000003` |
 | `PARITY_SHARE_NAMESPACE_ID`            | `NamespaceID` | `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF` |
 
+## Reserved State Subtree IDs
+
+| name                         | type             | value  |
+| ---------------------------- | ---------------- | ------ |
+| `ACCOUNTS_SUBTREE_ID`        | `StateSubtreeID` | `0x01` |
+| `VALIDATORS_SUBTREE_ID`      | `StateSubtreeID` | `0x02` |
```
Liamsi: In tendermint, the header references current and next validators:
https://github.com/tendermint/tendermint/blob/206c814a8e64cb4b9eb2abbb2fdadc6933b28584/types/block.go#L352-L353

Should we have two subtrees for that? Or, further split the validator subtree? The number of valitors should easily fit into that tree in any case.

specs/data_structures.md
```diff
@@ -496,19 +501,32 @@ One unified [Sparse Merkle Trees](#sparse-merkle-tree) is maintained for both ac
 | `balance`        | `uint64`                  | Coin balance.                                                                               |
 | `nonce`          | `uint64`                  | Account nonce. Every outgoing transaction from this account increments the nonce.           |
 | `isValidator`    | `bool`                    | Whether this account is a validator or not. Mutually exclusive with `isDelegating`.         |
-| `validatorInfo`  | [Validator](#validator)   | _Optional_, only if `isValidator` is set. Validator info.                                   |
 | `isDelegating`   | `bool`                    | Whether this account is delegating its stake or not. Mutually exclusive with `isValidator`. |
 | `delegationInfo` | [Delegation](#delegation) | _Optional_, only if `isDelegating` is set. Delegation info.                                 |
 
-In the accounts tree, accounts (i.e. leaves) are keyed by the [hash](#hashdigest) of their [address](#address).
+In the accounts subtree, accounts (i.e. leaves) are keyed by the [hash](#hashdigest) of their [address](#address). The first byte is then replaced with `ACCOUNTS_SUBTREE_ID`.
```
Liamsi, line 504: Effectively, this means that the hash used in the subtree returns 31 bytes.

specs/data_structures.md
```diff
@@ -540,27 +558,23 @@ Validator objects represent all the information needed to be keep track of a val
...
+| name            | type     | description                  |
+| --------------- | -------- | ---------------------------- |
+| `numValidators` | `uint32` | Number of active validators. |
 
-Delegation objects represent a delegation. They have two statuses:
-1. `Bonded`: This delegation is enabled for a `Queued` _or_ `Bonded` validator. Delegations to a `Queued` validator can be withdrawn immediately, while delegations for a `Bonded` validator must be unbonded first.
-1. `Unbonding`: This delegation is unbonding. It will remain in this status for at least `UNBONDING_DURATION` blocks, and while unbonding may still be slashed. Once the unbonding duration has expired, the delegation can be withdrawn.
+Since the [validator set](#validator) is stored in a Sparse Merkle Tree, there is no compact way of proving that the number of active validators exceeds `MAX_VALIDATORS` without keeping track of the number of active validators. There is only a single leaf in the active validator count subtree, which is keyed with zero (i.e. `0x0000000000000000000000000000000000000000000000000000000000000000`), and the first byte replaced with `VALIDATOR_COUNT_SUBTREE_ID`.
```
Liamsi: imo this should be part of validator subtree: 
https://github.com/lazyledger/lazyledger-specs/pull/32/commits/f1cba758138d074b7a18b59d62b3236587f920d0#r439773622

specs/data_structures.md
```diff
@@ -475,7 +481,101 @@ enum VoteType : uint8_t {
...
+1. `Unbonding`: This validator is in the process of unbonding, which can be voluntary (the validator decided to stop being an active validator) or forced (the validator committed a slashable offence and was kicked from the active validator set). Validators will remain in this status for at least `UNBONDING_DURATION` blocks, and while unbonding may still be slashed.
+1. `Unbonded`: This validator has completed its unbonding and has withdrawn its stake. The validator object will remain in this status until `delegatedCount` reaches zero, at which point it is destroyed.
+
+In the validators subtrees, validators are keyed by the [hash](#hashdigest) of their [address](#address). The first byte is then replaced with `ACTIVE_VALIDATORS_SUBTREE_ID` for the active validator set or `INACTIVE_VALIDATORS_SUBTREE_ID` for the inactive validator set. Active validators are `Bonded`, while inactive validators are not `Bonded`. By construction, the validators subtrees will be a subset of a mirror of the [accounts subtree](#account).
```
Liamsi, line 558: I don't really understand why we would need two trees here. Or why we need a tree for the inactive validator set.

specs/data_structures.md
```diff
@@ -66,16 +72,16 @@ Blocks are the top-level data structure of the LazyLedger blockchain.
...
+| name                | type                      | description                                                                  |
+| ------------------- | ------------------------- | ---------------------------------------------------------------------------- |
+| `height`            | `uint64`                  | Block height. The genesis block is at height `1`.                            |
+| `time`              | [Time](#time)             | Timestamp of this block.                                                     |
+| `lastBlockID`       | [BlockID](#blockid)       | Previous block's ID.                                                         |
+| `lastCommitRoot`    | [HashDigest](#hashdigest) | Previous block's Tendermint commit root.                                     |
+| `consensusRoot`     | [HashDigest](#hashdigest) | Merkle root of [consensus parameters](#consensus-parameters) for this block. |
+| `stateCommitment`   | [HashDigest](#hashdigest) | The [state root](#state) after this block's transactions are applied.        |
```
Liamsi, line 82: If it is clear that we will stick with deferred execution for a while (probably after launch even), should the first version of the spec adhere to this too? (then this would be state after the last block's transactions are applied).

Whenever (if) we decide to drop deferred execution, we can update the spec to another version.

specs/consensus.md
adlerjohn: It _could_, but mangling multiple leaf types into a single tree (or subtree) makes things more complicated. E.g. you'd need subtree-specific logic if you want to parallelize updating the validator set and the validator count state. If they're in separate subtrees then you can have a single global whole-SMT-level process that handles updating subtrees in parallel.

The validator count is needed not for light nodes knowing they've downloaded the entire validator set, but for proving with a fraud proof that the number of active validators exceeds the maximum.

specs/data_structures.md
adlerjohn: Yes.

specs/data_structures.md
adlerjohn: It's purely an optimization. Every time a delegation operation (delegating, undelegating) is made  on an inactive validator, an a validator moving to/from inactive and active, you have to remove the validator object from the active validator tree to somewhere else. If it's the accounts, then you have to do all those operations on the entire accounts subtree (larger tree and no parallelization). If it's in a dedicated inactive validators subtree, then that subtree will be smaller and computing the roots can be done in parallel trivially.

Not strictly necessary but more efficient and easier to implement in a more performant way. Remember that we need to serialize the state (i.e. compute the new state root) after every transaction for fraud proofs.

specs/data_structures.md
adlerjohn: We need immediate execution for fee burning (which IMO is 100% necessary, for testnet at the latest). Minting new coins is a source, and there are two ways to have sinks:
1. state rent
1. fee burning

We can't really do the former for obvious reasons, so we're left with the latter. Sinks create intrinsic demand for the currency, a property that cannot be accomplished any way else (which is why taxes---a sink---are still a thing even if the government can print money---a source). And fee burning requires immediate execution.

specs/consensus.md
adlerjohn: This isn't needed with immediate execution. Regardless, the validator set for the current block is the next validator set of the previous block, so we don't need to maintain two trees. Only the next validator set is needed to be stored in the tree at any given time.

specs/data_structures.md
adlerjohn: [Unlike moving the inactive validator set into the accounts subtree](#discussion_r439783834), I might be okay with moving the validator count into the active validators subtree because it's guaranteed to be in a fixed location.

specs/consensus.md
Liamsi: > Only the next validator set is needed to be stored in the tree at any given time.

That makes sense. I'm wondering if there is a reason in tendermint for having both validator sets referenced. Might have to do with deferred execution (related https://github.com/tendermint/tendermint/issues/2483) or maybe it is just for convenience 🤔

specs/data_structures.md
Liamsi: If we say it's the hash function, this is easier to formalize. e.g. we can say the key for an account in the set `Accs` tree is computed as:
`key := subtree_id || hash_st(address), `
where `hash_st` is any cryptographic hash function `hash_st: Accs -> {0,1}^31`

specs/data_structures.md
Liamsi: I think it semantically belongs to the validator subtree and also a single value tree seems weird.

specs/data_structures.md
Liamsi: Thanks for the explanantion!

specs/data_structures.md
Liamsi: I was under the impression, we will have a first test-net without switching to immediate execution. If we will use deferred execution only in dev-nets, it's probably OK that the spec and the implementation diverge for a while I guess. 

We might have to re-discuss this when we are closer to a first dev-net iteration.

adlerjohn: By testnet above I guess I meant "testnet with mainnet configuration." We could have more public testnets (e.g. PoA ones) before that, along with non-feature-complete devnets, that don't need immediate execution because there are no incentives and we can just assume the single operator isn't making invalid blocks.

specs/data_structures.md
adlerjohn: Alright, will migrate the count into the active validators subtree.

specs/data_structures.md
adlerjohn: Hmm. I'm not really a fan because that would mean we need 1) a different hashing function for each subtree (not great, not terrible) and 2) to slice off the first byte of every single hashing operation. That cost will add up quickly, especially if doing proofs in smart contracts (which will need to be done even in the absence of fraud proofs---Merkle proofs need to be provided for transaction and message inclusion to Ethereum contracts that use LL for data availability.

I'd much prefer just changing how the keys are calculated, which is a one-time calculation per leaf.

specs/data_structures.md
adlerjohn: Fixed in 4d87996.

specs/consensus.md
Liamsi: > Regardless, the validator set for the current block is the next validator set of the previous block,

How would a light client verify this property efficiently if only the next valset is stored & merkelized? Doesn't it need a commit to the next valst (of the previous block) that can be easyily verified on the current block without recomputing the cur vals?

specs/consensus.md
```diff
@@ -53,6 +60,15 @@ Consensus Rules
 | `EVIDENCE_NAMESPACE_ID`                | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000003` |
 | `PARITY_SHARE_NAMESPACE_ID`            | `NamespaceID` | `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF` |
 
+## Reserved State Subtree IDs
+
+| name                             | type             | value  |
+| -------------------------------- | ---------------- | ------ |
+| `ACCOUNTS_SUBTREE_ID`            | `StateSubtreeID` | `0x01` |
+| `ACTIVE_VALIDATORS_SUBTREE_ID`   | `StateSubtreeID` | `0x02` |
+| `INACTIVE_VALIDATORS_SUBTREE_ID` | `StateSubtreeID` | `0x03` |
```
Liamsi, line 69: Are these in relation to the current block or a commitment to the next block (i.e. the next block will be signed by the vals in ACTIVE_VALIDATORS_SUBTREE?

specs/data_structures.md
Liamsi: IMO It's just another way to writing down the same thing though (of course in 2) the slicing off could be done more efficiently by replacing the 1st byte instead; but this is just an implementation detail).

specs/consensus.md
adlerjohn: Transcluding conversation from Slack:
> Under immediate execution, the next validator set is actually committed to in the last intermediate state root's active validator set subtree.

It might be worth it to have a dedicated field in the block header for this, and just making sure it matches up with the last intermediate state root's active validator subtree root.

specs/consensus.md
adlerjohn: These represent the most current state after each and every transaction. But the state root committed to in the block header is after all transactions in the block have been applied (under immediate execution).

specs/consensus.md
Liamsi: @buchmann clarified here why both hashes are part of the header (and not just the state root/tree) in tendermint:
https://github.com/lazyledger/lazyledger-core/issues/3#issuecomment-644786246

PR #33, opened 2020-05-22
Add key to leaf hash for SMT

adlerjohn:
When hashing leaves for the SMT, also hash over the key of the leaf so that the key of each leaf is committed to.


specs/data_structures.md
```diff
@@ -277,12 +277,24 @@ Finally, the number of hashing operations can be reduced to be logarithmic in th
...
+
+For leaf node of leaf message `m` with key `k`, its value `v` is:
+```C++
+v = h(`0x00`, k, serialize(m))
+```
+
+For internal node with children `l` and `r`, its value `v` is:
+```C++
+v = h(`0x01`, l.v, r.v)
+```
```
Liamsi, line 290: 👍

PR #37, opened 2020-06-20
Change protobuf timestamps to TAI64 timestamps

adlerjohn:
Remove dependence on protobuf for core data structures. [TAI64](http://cr.yp.to/libtai/tai64.html) is chosen as the time format both because it's reasonable (unsigned 64 bit integer) and because it's the same format used by [Veriform](https://github.com/iqlusioninc/veriform).

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -111,7 +111,7 @@ Data that is [erasure-coded](#erasure-coding) for [data availability checks](htt
 
 ## Time
 
-LazyLedger uses the [Google Protobuf Timestamp](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp) format for timestamps, which represents time as seconds in UTC Epoch Time and nanoseconds.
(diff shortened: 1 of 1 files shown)


specs/data_structures.md
```diff
@@ -111,7 +111,7 @@ Data that is [erasure-coded](#erasure-coding) for [data availability checks](htt
 
 ## Time
 
-LazyLedger uses the [Google Protobuf Timestamp](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp) format for timestamps, which represents time as seconds in UTC Epoch Time and nanoseconds.
+LazyLedger uses a 64-bit unsigned integer (`uint64`) to represent time in [TAI64](http://cr.yp.to/libtai/tai64.html) format.
```
liamsi, line 114: Previously, this was more self-contained (as it stated "time as seconds in UTC Epoch Time and nanoseconds"). Now, one has to go to another site, to understand this are seconds.

adlerjohn: That is true, but at the same time, it turns out that "Epoch time" is non-trivial to define. You need to account for leap seconds, and every scheme has whole paragraphs around that. I found it better to just describe how the data is stored, and leave defining what goes in that integer to the time spec.

PR #42, opened 2020-06-29
Add transaction types

adlerjohn:
- [x] Add basic transaction types.
- [x] Add new namespace ID for tail padding (i.e. padding after the last real message).

Rendered:
- https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-transaction_types/specs/consensus.md
- https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-transaction_types/specs/data_structures.md


specs/consensus.md
```diff
@@ -58,6 +58,7 @@ Consensus Rules
 | `TRANSACTION_NAMESPACE_ID`             | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000001` |
 | `INTERMEDIATE_STATE_ROOT_NAMESPACE_ID` | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000002` |
 | `EVIDENCE_NAMESPACE_ID`                | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000003` |
+| `TAIL_PADDING_NAMESPACE_ID`            | `NamespaceID` | `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE` |
```
liamsi, line 61: Shouldn't these values have a brief description too? 
It might be unclear what tail padding is without explaining "padding after the last real message".

specs/data_structures.md
```diff
@@ -406,16 +416,153 @@ Wrapped transactions include additional metadata by the block proposer that is c
...
+
+#### SignedTransactionData: PayForPadding
+
+| name                 | type                                       | description                                                  |
+| -------------------- | ------------------------------------------ | ------------------------------------------------------------ |
+| `type`               | `TransactionType`                          | Must be `TransactionType.PayForPadding`.                     |
+| `messageNamespaceID` | [`NamespaceID`](consensus.md#type-aliases) | Namespace ID of padding this transaction pays a fee for.     |
+| `messageSize`        | `uint64`                                   | Size of padding this transaction pays a fee for, in `byte`s. |
+
+Pays for the inclusion of a padding shares in the same block. Padding shares are used between real messages that are not tightly packed. For rationale, see [rationale doc](../rationale/message_block_layout.md).
```
liamsi, line 501: Will users send two tx then? one for the inclusion of the message and an optional extra one for padding only?

adlerjohn: The transactions here are based on the staking messages of the Cosmos SDK: https://docs.cosmos.network/master/modules/staking/03_messages.html. Some features are removed (one-transaction re-delagation, for example, since it adds protocol complexity that can be abstracted away at the wallet level).

adlerjohn:
> Left one minor comment. I'd like to compare the tx types to the particular messages (this word is heavily overloaded in the different contexts we use it)/ Tx in the relevant cosmos sdk modules but this does not block merging this PR.

The transactions here are based on the staking messages of the Cosmos SDK: https://docs.cosmos.network/master/modules/staking/03_messages.html. Some features are removed (one-transaction re-delagation, for example, since it adds protocol complexity that can be abstracted away at the wallet level).

specs/data_structures.md
adlerjohn: Not necessarily (i.e. no, but subject to change). Interstitial padding should be paid (or, at least, _claimed_) by the block proposer IMO.

specs/consensus.md
adlerjohn: Fixed in 0abd950.

specs/data_structures.md
adlerjohn: Clarified in c001036.

specs/consensus.md
```diff
@@ -53,13 +53,13 @@ Consensus Rules
...
+| -------------------------------------- | ------------- | -------------------------------------------------------------------- | ----------------------------------------------------------------------------- |
+| `TRANSACTION_NAMESPACE_ID`             | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000001` | Transactions: requests that modify the state.                                 |
+| `INTERMEDIATE_STATE_ROOT_NAMESPACE_ID` | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000002` | Intermediate state roots, committed after every transaction.                  |
+| `EVIDENCE_NAMESPACE_ID`                | `NamespaceID` | `0x0000000000000000000000000000000000000000000000000000000000000003` | Evidence: fraud proofs or other proof of slashable action.                    |
+| `TAIL_PADDING_NAMESPACE_ID`            | `NamespaceID` | `0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFE` | Tail padding: padding after all messages to fill up the original data square. |
```
liamsi, line 61: Because of the 32 bytes this doesn't look nice when rendered. I wonder if we should write the namespaces as 1, 2, 3, 2^256-2, 2^256-1 respectively and mention that they are represented as 32 bytes.

adlerjohn: Hmmm. It people view it on https://docs.lazyledger.io/lazyledger-specs then we can configure the CSS however we want, so this might be a non-issue?

liamsi: Yes, let's see how it looks there 👍 In general 1, 2, 3, 2^256-2, 2^256-1 might be easier for a human to parse and can still be used to init a 32 byte array correctly when implementing. Let's keep it as is for now.

Issue #41, opened 2020-06-24
Merkle trees misc cleanup and fixes


liamsi:
https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#sparse-merkle-tree

> For leaf node of leaf message m with key k, its value v is:

Better to use a different word than message here (otherwise this might be conflated with the other use of the word  [messages](https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#message))

Also, it would help getting a quick understanding, if there was 2-3 simple to follow examples how an smt and its will look like (including trivial edge-case like all-empty tree, only one entry, and one realistic example with a few entries).

adlerjohn:
Things to add:
- Split up computing the root and Merkle proofs into different subsections
- Example diagrams (especially one that shows that internal nodes w/single child are possible)
- Define zero tree base case
- Fix wording on leaf messages

liamsi:
> The annotated Merkle tree abstraction isn't suitable for prefixing to differentiate between internal and leaf nodes. Need to make that an explicit feature.

Independent of the prefixing, I'm not sure the annotated and the namespaced merkle tree hashes the correct thing.

Looking at how @musalbas defined the hashing in his paper:
> We define a wrapper function nsHash(x) that produces hashes  prefixed with namespace identifiers. A namespaced hash has the format `minNs||maxNs||hash(x)`, where `minNs` is the lowest namespace identifier found in all the children of the node that the hash represents, and `maxNs` is the highest.

OK, and:

> If x is a leaf, then minNs = maxNs = ns(x), as the hash contains only one leaf with a single namespace

Using the notation in the spec, my understanding is that this would result in:
```
n_min = m_1_l(m) = m.namespace_id
n_max = m_2_l(m) = m.namespace_id
v = n_min || n_max || h(serialize(m))
```
but we have:
```
n_min = m_1_l(m) = m.namespace_id
n_max = m_2_l(m) = m.namespace_id
v = h(serialize(m))
```
https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#namespace-merkle-tree

Similarly, for the inner nodes, I think the hash of should be something like:
`v = nsh(l, r) = l.n_min || l.n_max || h( l.v,  r.v )` 
instead of currently:
`v = h(l, r) = h(l.n_min, l.n_max, l.v, r.n_min, r.n_max, r.v)`

[...]

PR #45, opened 2020-07-01
Add share namespace ID field

adlerjohn:
- Add namespace ID field to share data structure.
- Add note on serialization of shares to only use raw data.

Note: I intentionally did not fix the conflicting use of "message" for NMTs. That'll be handled in #41.

Rendered:
- https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-share_namespace/specs/data_structures.md

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -29,6 +29,7 @@ Data Structures
   - [Reed-Solomon Erasure Coding](#reed-solomon-erasure-coding)
   - [2D Reed-Solomon Encoding Scheme](#2d-reed-solomon-encoding-scheme)
   - [Share](#share)
+    - [Share Serialization](#share-serialization)
   - [Arranging Available Data Into Shares](#arranging-available-data-into-shares)
 - [Available Data](#available-data)
   - [TransactionData](#transactiondata)
@@ -277,12 +278,12 @@ The NMT is an annotated Merkle tree with two additional fields and methods that 
 
 For leaf node of message `m`:
 ```C++
-n_min = m_1_l(m) = m.namespace_id
-n_max = m_2_l(m) = m.namespace_id
+n_min = m_1_l(m) = m.namespaceID
+n_max = m_2_l(m) = m.namespaceID
 v = h(serialize(m))
 ```
 
-The `namespace_id` message field here is the namespace ID of the message, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-byte unsigned integer.
+The `namespaceID` message field here is the namespace ID of the message, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-byte unsigned integer.
 
 Before being hashed, the [messages](#message) are [serialized](#serialization).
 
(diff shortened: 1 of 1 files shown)


specs/data_structures.md
```diff
@@ -379,16 +380,21 @@ Finally, the `availableDataRoot` of the block [Header](#header) is computed as t
 
 A share is a fixed-size data chunk that will be erasure-coded and committed to in [Namespace Merkle trees](#namespace-merkle-tree).
 
-| name      | type               | description     |
-| --------- | ------------------ | --------------- |
-| `rawData` | `byte[SHARE_SIZE]` | Raw share data. |
+| name          | type                         | description                |
+| ------------- | ---------------------------- | -------------------------- |
+| `namespaceID` | [NamespaceID](#type-aliases) | Namespace ID of the share. |
```
liamsi, line 385: A LL message is associated with a namespaceID. What is the namespace of a share though? Can't a share contain the end of one message and the beginning of another one (with different NIDS)?

specs/data_structures.md
```diff
@@ -379,16 +380,21 @@ Finally, the `availableDataRoot` of the block [Header](#header) is computed as t
...
 
 An example layout of the share's internal bytes is shown below. For non-parity shares _with a reserved namespace_, the first `SHARE_RESERVED_BYTES` bytes (`*` in the figure) is the starting byte of the first request in the share as an unsigned integer, or `0` if there is none. In this example, the first byte would be `80` (or `0x50` in hex). For shares _with a non-reserved namespace_ (and parity shares), the first `SHARE_RESERVED_BYTES` bytes have no special meaning and are simply used to store data like all the other bytes in the share.
 
 ![fig: Reserved share.](./figures/share.svg)
 
 For non-parity shares, if there is insufficient request data to fill the share, the remaining bytes are padded with `0`.
 
+### Share Serialization
+
+Shares [canonically serialized](#serialization) using only the raw share data, i.e. `serialize(share) = serialize(share.rawData)`.
```
liamsi, line 396: Just for my understanding: In the context of the NMT this means, that a leaf (share) will get hashed in the following way:
`nsid||nsid||hash(leafPrefix||leaf)`, where `leaf =  raw_data` 
as opposed to:
`nsid||nsid||hash(leafPrefix||leaf)`, where `leaf = nid || raw_data`

https://github.com/liamsi/nmt-experiments/blob/70b02631b0e79986ec266b514e6da607e5c8e9d7/trillian_based/hasher.go#L48

specs/data_structures.md
adlerjohn: No. From the rationale doc (https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-share_namespace/rationale/message_block_layout.md):

> 1. Data must be ordered by namespace ID. This makes queries into a NMT commitment of that data more efficient.
> 2. Since non-message data are not naturally intended for particular namespaces, we assign reserved namespaces for them. A range of namespaces is reserved for this purpose, starting from the lowest possible namespace ID.
> 3. By construction, the above two rules mean that non-message data always precedes message data in the row-major matrix, even when considering single rows or columns.
> 4. Data with different namespaces must not be in the same share. This might cause a small amount of wasted block space, but makes the NMT easier to reason about in general since leaves are guaranteed to belong to a single namespace.

By (4) shares will only have a single namespace ID associated with them. The reason shares have namespace IDs is that the NMT has shares as its leaves, and we need to know the NIDs of each leaf (i.e. share).

specs/data_structures.md
adlerjohn: Right.

specs/data_structures.md
adlerjohn: To expand on that, the NMT doesn't care if it's hashing messages or transactions or evidence. It just sees shares. And each share needs to have a NID somehow.

liamsi: I see. And the order for account balance related Tx is preserved because they end up in the right order in their preserved namespace?

PR #46, opened 2020-07-01
Add new burn tx type

adlerjohn:
Add a new type of tx, `Burn`, which burns some coins. This can be used for one-way bridges (e.g. burning coins then minting them in a LazyLedger app). The burn tx includes a graffiti field to specify the burn reason.

Rendered:
- https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-burn_tx/specs/data_structures.md

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -45,6 +45,7 @@ Data Structures
       - [SignedTransactionData: CreateDelegation](#signedtransactiondata-createdelegation)
       - [SignedTransactionData: BeginUnbondingDelegation](#signedtransactiondata-beginunbondingdelegation)
       - [SignedTransactionData: UnbondDelegation](#signedtransactiondata-unbonddelegation)
+      - [SignedTransactionData: Burn](#signedtransactiondata-burn)
   - [IntermediateStateRootData](#intermediatestaterootdata)
     - [WrappedIntermediateStateRoot](#wrappedintermediatestateroot)
     - [IntermediateStateRoot](#intermediatestateroot)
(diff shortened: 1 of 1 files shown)


specs/data_structures.md
```diff
@@ -75,6 +76,7 @@ Data Structures
 | `Amount`                    | `uint64`                   |
 | [`BlockID`](#blockid)       | [HashDigest](#hashdigest)  |
 | `FeeRate`                   | `uint64`                   |
+| `Graffiti`                  | `NamespaceID`              |
```
liamsi, line 79: > The burn tx includes a graffiti field to specify the burn reason.

Sounded more like a free text memo field btw. why not call the field "Namespace" instead of graffiti?

specs/data_structures.md
```diff
@@ -584,6 +587,16 @@ Begin unbonding the [Delegation](#delegation) at this address.
...
 
+#### SignedTransactionData: Burn
+
+| name         | type                      | description                                  |
+| ------------ | ------------------------- | -------------------------------------------- |
+| `type`       | `TransactionType`         | Must be `TransactionType.Burn`.              |
+| `amount`     | [Amount](#type-aliases)   |                                              |
+| `maxFeeRate` | [FeeRate](#type-aliases)  |                                              |
+| `nonce`      | [Nonce](#type-aliases)    |                                              |
+| `graffiti`   | [Graffiti](#type-aliases) | Graffiti to indicate the reason for burning. |
```
liamsi, line 598: related: https://github.com/lazyledger/lazyledger-specs/pull/46/files#r448381336

specs/data_structures.md
adlerjohn: So, my reasoning for having a separate graffiti type is that it's not 100% guaranteed that burning will be targeted to a particular namespace. It could be anything! Using namespace ID directly would introduce a tight coupling. If we decide to shrink the namespace ID later, then we'd have to change the field in the transaction from namespace ID to e.g. hash digest. With a separate graffiti type alias, we can just change the type being aliased.

liamsi: > With a separate graffiti type alias, we can just change the type being aliased.

which is effectively the same as changing the field (has the same implications at least).

PR #47, opened 2020-07-02
Merkle tree misc fixes

adlerjohn:
- [x] Change binary Merkle tree to use scheme in [RFC 6962](https://tools.ietf.org/html/rfc6962), i.e. unbalanced tree.
- [x] Split up computing the root and Merkle proofs into different subsections
- [ ] Example diagrams. (#49)
- [x] Define zero tree base case.
- [x] Fix wording on leaf messages -> leaf data.
- [ ] Merkle multiproof format. (#48)
- [x] Remove the annotated Merkle tree abstraction since it isn't used elsewhere in the spec except for NMT. Define the domain separation prefixing for base Merkle trees.

Rendered:
- https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-merkle_tree_fixes/specs/data_structures.md

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -251,6 +251,13 @@ Leaves and internal nodes are hashed differently: the one-byte `0x00` is prepend
 
 [Shares](#share) in LazyLedger are associated with a provided _namespace ID_. The Namespace Merkle Tree (NMT) is a variation of the [Merkle Interval Tree](https://eprint.iacr.org/2018/642), which is itself an extension of the [Merkle Sum Tree](https://bitcointalk.org/index.php?topic=845978.0). It allows for compact proofs around the inclusion of exclusion of shares with particular namespace IDs.
 
+The base case (an empty tree) is defined as:
+```C++
+n_min = 0x0000000000000000000000000000000000000000000000000000000000000000
+n_max = 0x0000000000000000000000000000000000000000000000000000000000000000
+v = 0x0000000000000000000000000000000000000000000000000000000000000000
+```
+
 For leaf node of data `d`:
 ```C++
 n_min = d.namespaceID
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -231,6 +231,11 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
 
 Binary Merkle trees are constructed in the same fashion as described in [Certificate Transparency (RFC-6962)](https://tools.ietf.org/html/rfc6962). Leaves are hashed once to get leaf node values and internal node values are the hash of the concatenation of their children (either leaf nodes or other internal nodes).
 
+Nodes contain a single field:
+| name | type       | description |
+| ---- | ---------- | ----------- |
+| `v`  | `byte[32]` | Node value. |
+
 The base case (an empty tree) is defined as zero:
 ```C++
 node.v = 0x0000000000000000000000000000000000000000000000000000000000000000
@@ -262,6 +267,13 @@ Leaves and internal nodes are hashed differently: the one-byte `0x00` is prepend
 ## Namespace Merkle Tree
 
 [Shares](#share) in LazyLedger are associated with a provided _namespace ID_. The Namespace Merkle Tree (NMT) is a variation of the [Merkle Interval Tree](https://eprint.iacr.org/2018/642), which is itself an extension of the [Merkle Sum Tree](https://bitcointalk.org/index.php?topic=845978.0). It allows for compact proofs around the inclusion or exclusion of shares with particular namespace IDs.
+
+Nodes contain three fields:
+| name    | type                         | description                                      |
+| ------- | ---------------------------- | ------------------------------------------------ |
+| `n_min` | [NamespaceID](#type-aliases) | Min namespace ID in subtree rooted at this node. |
+| `n_max` | [NamespaceID](#type-aliases) | Max namespace ID in subtree rooted at this node. |
+| `v`     | `byte[32]`                   | Node value.                                      |
 
 The base case (an empty tree) is defined as:
 ```C++
@@ -305,6 +317,11 @@ Additional rules are added on top of plain [binary Merkle trees](#binary-merkle-
 1. Default values are given to leaf nodes with empty leaves.
 1. While the above rule is sufficient to pre-compute the values of intermediate nodes that are roots of empty subtrees, a further simplification is to extend this default value to all nodes that are roots of empty subtrees. The 32-byte zero, i.e. `0x0000000000000000000000000000000000000000000000000000000000000000`, is used as the default value. This rule takes precedence over the above one.
 1. The number of hashing operations can be reduced to be logarithmic in the number of non-empty leaves on average, assuming a uniform distribution of non-empty leaf keys. An internal node that is the root of a subtree that contains exactly one non-empty leaf is replaced by that leaf's leaf node.
+
+Nodes contain a single field:
+| name | type       | description |
+| ---- | ---------- | ----------- |
+| `v`  | `byte[32]` | Node value. |
 
 The base case (an empty tree) is defined as the default value:
 ```C++
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -21,7 +21,9 @@ Data Structures
 - [Public-Key Cryptography](#public-key-cryptography)
 - [Merkle Trees](#merkle-trees)
   - [Binary Merkle Tree](#binary-merkle-tree)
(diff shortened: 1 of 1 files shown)


adlerjohn:
I think I'll relegate making nice diagrams and the Merkle multiproof format to issues for later, since they're not critical atm.

specs/data_structures.md
```diff
@@ -21,10 +21,11 @@ Data Structures
 - [Public-Key Cryptography](#public-key-cryptography)
 - [Merkle Trees](#merkle-trees)
   - [Binary Merkle Tree](#binary-merkle-tree)
-  - [Annotated Merkle Tree](#annotated-merkle-tree)
-    - [Verifying Annotated Merkle Proofs](#verifying-annotated-merkle-proofs)
```
liamsi, line 25: 👍

specs/data_structures.md
```diff
@@ -228,100 +229,107 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
...
+n_max = 0x0000000000000000000000000000000000000000000000000000000000000000
+v = 0x0000000000000000000000000000000000000000000000000000000000000000
 ```
 
-The `namespaceID` message field here is the namespace ID of the message, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-long byte array.
+For leaf node of data `d`:
+```C++
+n_min = d.namespaceID
+n_max = d.namespaceID
+v = h(0x00, serialize(m))
```
liamsi, line 277:
```suggestion
v = h(0x00, serialize(d)
```
?

liamsi: Also, it's not clear that `n_min` and `n_max` belong to the leaf now. 

Another less ambiguous way to write this down is:
```suggestion
v = n_min||n_max||h(0x00, serialize(d))
```
Then for the internal node case you could write
`v = h(l, r) = n_min||n_max||h(0x01, l, r)`

specs/data_structures.md
```diff
@@ -228,100 +229,107 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
...
+The `namespaceID` message field here is the namespace ID of the leaf, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-long byte array.
 
 For internal node with children `l` and `r`:
 ```C++
-n_min = m_1_i(height, l, r) = min(l.n_min, r.n_min)
-n_max = m_2_i(height, l, r) = max(l.n_max, r.n_max)
-v = h(l, r) = h(l.n_min, l.n_max, l.v, r.n_min, r.n_max, r.v)
+n_min = min(l.n_min, r.n_min)
+n_max = max(l.n_max, r.n_max)
+v = h(l, r) = h(0x01, l.n_min, l.n_max, l.v, r.n_min, r.n_max, r.v)
```
liamsi, line 286: Isn't the `n_min||n_max||` prefix missing here?

specs/data_structures.md
```diff
@@ -228,100 +229,107 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
...
+| name            | type                             | description                                                              |
+| --------------- | -------------------------------- | ------------------------------------------------------------------------ |
+| `root`          | [HashDigest](#hashdigest)        | Merkle root.                                                             |
+| `key`           | `byte[32]`                       | Key (i.e. index) of the leaf.                                            |
+| `depth`         | `uint16`                         | Depth of the leaf node. The root node is at depth `0`. Must be `<= 256`. |
+| `siblingValues` | [HashDigest](#hashdigest)`[]`    | Sibling hash values.                                                     |
+| `siblingMins`   | [NamespaceID](#type-aliases)`[]` | Sibling min namespace IDs.                                               |
+| `siblingMaxes`  | [NamespaceID](#type-aliases)`[]` | Sibling max namespace IDs.                                               |
+| `leaf`          | `byte[]`                         | Leaf value.                                                              |
```
liamsi: 👍

specs/data_structures.md
liamsi: It's unclear the `n_min` and `n_max` become part of the node here.

adlerjohn: No, these two are isomorphisms:
- tuple of byte arrays `<n_min, n_max, v>`
- single byte array `n_min || n_max || v`

liamsi: > No, these two are isomorphisms:

sure thing, and I understand what you mean here. Still this will be confusing because that part:
> I'm using the first representation here (i.e. a node has 3 fields: node.n_min, node.n_max and node.v).

is kinda missing in the spec.

liamsi: If you mean to say a node has 3 fields, then it should be mentioned somewhere in the spec. and regarding 
https://github.com/lazyledger/lazyledger-specs/pull/47#discussion_r449089619
those should then be `v.n_min` and `v.n_max` respectively.

adlerjohn: Should be fixed as of 16e468d.

specs/data_structures.md
```diff
@@ -228,100 +229,129 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
...
+node.n_max = 0x0000000000000000000000000000000000000000000000000000000000000000
+node.v = 0x0000000000000000000000000000000000000000000000000000000000000000
 ```
 
-The `namespaceID` message field here is the namespace ID of the message, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-long byte array.
+For leaf node `node` of data `d`:
+```C++
+node.n_min = d.namespaceID
+node.n_max = d.namespaceID
+node.v = h(0x00, serialize(d))
```
liamsi, line 289: One last nitpick: You implicitly assuming that data `d` is a structure as well namely `struct{namespaceID, raw_data}`. And that last line is still ambiguous to as if `h(0x00, d.raw_data)` or `h(0x00, d.namespaceID||d.raw_data))` (or whatever the result of `serialize(d)` is).

adlerjohn: Technically, the serialization of shares is defined as a special case: https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-merkle_tree_fixes/specs/data_structures.md#share-serialization.

> Shares canonically serialized using only the raw share data, i.e. `serialize(share) = serialize(share.rawData)`.

liamsi: Ah I missed that, That doesn't address that you implicitly assume that `d` has to have a structure (namely two fields) but I guess the important part is covered in the spec (how the shares end up in the tree ...).

specs/data_structures.md
```diff
@@ -228,100 +229,129 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
 
 ## Binary Merkle Tree
 
-Binary Merkle trees are constructed in the usual fashion, with leaves being hashed once to get leaf node values and internal node values being the hash of the concatenation of their children. The specific mechanism for hashing leaves for leaf nodes and children for internal nodes may be different (see: [annotated Merkle trees](#annotated-merkle-tree)), but for plain binary Merkle trees are the same.
+Binary Merkle trees are constructed in the same fashion as described in [Certificate Transparency (RFC-6962)](https://tools.ietf.org/html/rfc6962). Leaves are hashed once to get leaf node values and internal node values are the hash of the concatenation of their children (either leaf nodes or other internal nodes).
```
musalbas, line 232: It would be helpful to specify how it's the same as RFC 6962. Specifically, that the tree is unbalanced when there isn't 2^x leaves. The next sentence is describing properties of all Merkle trees.

specs/data_structures.md
```diff
@@ -228,100 +229,129 @@ Merkle trees are used to authenticate various pieces of data across the LazyLedg
...
 ```C++
-f_1 = m_1_i(height, l.f_1, r.f_1)
-...
-f_n = m_n_i(height, l.f_n, r.f_n)
-v = h(l.f_1, ..., l.f_n, l.v, r.f_1, ..., r.f_n, r.v)
+node.v = h(0x01, serialize(l), serialize(r))
 ```
 
-If a compact Merkle root is needed, the root level (which consists of root fields and a root value) can be hashed once.
+Note that rather than duplicating the last node if there are an odd number of nodes (the [Bitcoin design](https://github.com/bitcoin/bitcoin/blob/5961b23898ee7c0af2626c46d5d70e80136578d3/src/consensus/merkle.cpp#L9-L43)), trees are allowed to be imbalanced. In other words, the height of each leaf may be different. For an example, see Section 2.1.3 of [Certificate Transparency (RFC-6962)](https://tools.ietf.org/html/rfc6962).
```
musalbas, line 254: Ah nvm, it's specified here.

PR #52, opened 2020-07-24
Cleanup markdown link checker

adlerjohn:
Minor fixes:
- [x] Use checkout v2 instead of master
- [x] Rename action config in case we need more actions later


.github/workflows/markdown_link_checker.yml
```diff
@@ -6,5 +6,5 @@ jobs:
   markdown-link-check:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@v2
```
liamsi, line 9: Not sure what the difference between both is but probably a good idea to use a released version.

PR #55, opened 2020-07-27
Use 32 byte addresses

adlerjohn:


musalbas:
LGTM. I assume this would still allow for Ethereum compatibility, as the address can be truncated by the Ethereum smart contract?

adlerjohn:
Right, recovering the address of a signature on Ethereum will only use 20 bytes, and so only have 160 bits of security. But all other operations can just use the full 32 bytes no problem.

PR #59, opened 2020-08-04
Ignore namespace ID for parity shares

adlerjohn:
Ignore the namespace ID for siblings in the NMT if the sibling is the root of only parity shares, unless both siblings are that.

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -290,6 +290,8 @@ node.v = h(0x00, serialize(d))
 
 The `namespaceID` message field here is the namespace ID of the leaf, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-long byte array.
 
+Leaves in an NMT **must** be lexicographically sorted by namespace ID in ascending order.
+
 For internal node `node` with children `l` and `r`:
 ```C++
 node.n_min = min(l.n_min, r.n_min)
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -303,7 +303,9 @@ else
   node.n_max = max(l.n_max, r.n_max)
 node.v = h(l, r) = h(0x01, serialize(l), serialize(r))
 ```
-_Note that the above snippet leverages the property that leaves are sorted by namespace ID._ In plain English, the min and max namespace IDs for subtree roots with at least one non-parity leaf (which includes the root of an NMT, as [the right half of an NMT as used in LazyLedger will be parity shares](#2d-reed-solomon-encoding-scheme)) ignore the namespace ID for the parity leaves. Subtree roots with _only parity leaves_ have their min and max namespace ID set to [`PARITY_SHARE_NAMESPACE_ID`](consensus.md#reserved-state-subtree-ids).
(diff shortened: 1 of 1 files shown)


musalbas:
It doesn't look to me line the code snippet matches the text written for it?

adlerjohn:
The first condition should, in full, be checking if `l.n_min == PARITY_SHARE_NAMESPACE_ID && r.n_min == PARITY_SHARE_NAMESPACE_ID`, but the second condition can be deduplicated since leaves must be sorted by namespace ID. I don't actually point out explicitly that leaves must be sorted by namespace ID, so I probably should!

specs/data_structures.md
```diff
@@ -290,6 +290,8 @@ node.v = h(0x00, serialize(d))
 
 The `namespaceID` message field here is the namespace ID of the leaf, which is a [`NAMESPACE_ID_BYTES`](consensus.md#system-parameters)-long byte array.
 
+Leaves in a NMT **must** be lexicographically sorted by namespace ID in ascending order.
```
musalbas, line 293: I think it should be "an NMT"

specs/data_structures.md
```diff
@@ -301,7 +303,7 @@ else
   node.n_max = max(l.n_max, r.n_max)
```
musalbas, line 303: I think the code snippet would be clearer if it mention that only the min namespace IDs are checked because if the min is a parity share, then the max should be too, and why.

specs/data_structures.md
adlerjohn: Fixed in 19a80df.

specs/data_structures.md
adlerjohn: I could just change it to comparing `n_max`?

specs/data_structures.md
musalbas: https://blog.apastyle.org/apastyle/2012/04/using-a-or-an-with-acronyms-and-abbreviations.html

NMT is pronounced "en em tee", thus should start be prefixed with "an".

specs/data_structures.md
musalbas: But then it wouldn't be clear with n_max is compared but not n_min?

adlerjohn: Clarified in 455581a.

musalbas:
Is it actually mentioned anywhere in the spec apart from the intuition paragraph that subtree roots with only parity leaves have their min and max namespace ID set to `PARITY_SHARE_NAMESPACE_ID`?

adlerjohn:
> [Is it actually mentioned anywhere in the spec apart from the intuition paragraph that subtree roots with only parity leaves have their min and max namespace ID set to PARITY_SHARE_NAMESPACE_ID?](https://github.com/lazyledger/lazyledger-specs/pull/59#issuecomment-669271680)

No, and it's not really a relevant property anywhere else. It might up being brought up again when talking about application proofs, but that doesn't have much to do with consensus, it's more of a network protocol thing.

musalbas:
Shouldn't it be mentioned outside of the intuition paragraph? It seems like the base/special case of the namespace min/max calculation function. Putting it inside the intuition paragraph makes it seem like it's just a sidenote.

adlerjohn:
There's a line `By construction, either both the min and max namespace IDs of a node will be PARITY_SHARE_NAMESPACE_ID, or neither will`, which says the same thing. The explanation right after the snippet and before the intuition provides a more rigorous explanation. The intuition paragraph doesn't introduce any property that can't be derived from the previous paragraph.

PR #61, opened 2020-09-03
Add release drafter CI.

adlerjohn:
Add a CI task to automatically create release drafts with a changelog of PRs.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,19 @@
+name-template: "v$RESOLVED_VERSION"
+tag-template: "v$RESOLVED_VERSION"
+change-template: "- $TITLE (#$NUMBER) @$AUTHOR"
+change-title-escapes: '\<*_&'
+version-resolver:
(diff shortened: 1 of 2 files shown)


musalbas:
Is there any way to make this pull the submodule at https://github.com/lazyledger/docs.lazyledger.io too?

liamsi (approved):
I'm not familiar with the release-drafter/release-drafter action but given that this seem to work fine in the solidity proto plugin, this LGTM.

liamsi:
To @musalbas question, I think it's at least possible using this event: https://docs.github.com/en/developers/webhooks-and-events/webhook-events-and-payloads#release
Probably there is another action that can trigger an upload (deployment) after a release is published.

adlerjohn:
@musalbas you can add an action like [this](https://github.com/lazyledger/protobuf3-solidity-lib/blob/master/.github/workflows/npm-publish.yml) to run a command (with a secret token) when a release is published.

Issue #62, opened 2020-09-07
Should we decrease NAMESPACE_ID_BYTES?


liamsi:
Currently `NAMESPACE_ID_BYTES == 32` see: 
- https://github.com/lazyledger/lazyledger-specs/blob/master/specs/consensus.md#constants
- https://github.com/lazyledger/lazyledger-specs/blob/master/specs/consensus.md#reserved-namespace-ids

Just spitballing here but I think 16, or even just 8 bytes should be enough. 

ref: https://github.com/lazyledger/nmt/pull/10#discussion_r484563453

adlerjohn:
The namespace ID must be present twice (mix and max) for every since row root, and possibly every single column root, in the non-compact block header. It must also be present once per level in Merkle proofs into a NMT. In the happy case, the primary overhead is in the row and column roots.

Each root is a 32-bytes hash plus two namespace IDs. If namespace IDs are 32 bytes each, that in total triples the size of row and column roots. Back-of-the-napkin math puts Tendermint commit data around 5kb, and the row and column roots (just the roots) at around the same for a 1MB block. So adding in 32-byte namespace IDs would ~double the block header size. Definitely non-trivial.

There are really two reasonable possibilities:
- 32 bytes
  - This allows using a hash function's output directly, so we can just hash some identifier (e.g. `hash("Virtual sidechain 1")`) to get a unique namespace ID without a centralized registry. This actually isn't amazingly useful because it requires knowing that the namespace should be "Virtual sidechain 1" and not "virtual sidechain 1", for example. So in practice some registry would still be needed.
- 8 bytes
  - This is more than enough for sequential indices until the heat death of the universe (4 bytes, not so much). It does require a registry for mapping indices to some network, but as above that's ultimately a requirement if using 32 bytes as well.

So overall I think moving to 8 bytes per namespace ID reduces block header sizes non-trivially, and doesn't have any meaningful downsides to using 32 bytes.

musalbas:
I think even with 8 byte namespace, you can use the hash function output directly truncated to 8 bytes. Collision-resistance is not needed. Someone creating a virtual sidechain identifier that has the same hash as another virtual sidechain is no different than someone submitting spam transactions to a virtual sidechain.

PR #70, opened 2020-09-15
Reduce namespace ID size to 8 bytes, from 32

adlerjohn:
- Reduce size of namespace ID to 8 bytes, from 32 ([rationale](https://github.com/lazyledger/lazyledger-specs/issues/62#issuecomment-692894278))
- Unlink graffiti from namespace ID type definition

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -76,7 +76,7 @@ Data Structures
 | `Amount`                    | `uint64`                   |
 | [`BlockID`](#blockid)       | [HashDigest](#hashdigest)  |
 | `FeeRate`                   | `uint64`                   |
-| `Graffiti`                  | `NamespaceID`              |
+| `Graffiti`                  | `byte[GRAFFITI_BYTES]`     |
 | [`HashDigest`](#hashdigest) | `byte[32]`                 |
 | `Height`                    | `uint64`                   |
 | `NamespaceID`               | `byte[NAMESPACE_ID_BYTES]` |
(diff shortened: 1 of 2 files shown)


liamsi:
Cool, in the branch I'm working on (and haven’t pushed yet), I silently assumed 8 bytes too already 👍

specs/data_structures.md
```diff
@@ -76,7 +76,7 @@ Data Structures
 | `Amount`                    | `uint64`                   |
 | [`BlockID`](#blockid)       | [HashDigest](#hashdigest)  |
 | `FeeRate`                   | `uint64`                   |
-| `Graffiti`                  | `NamespaceID`              |
+| `Graffiti`                  | `byte[GRAFFITI_BYTES]`     |
```
liamsi, line 79: Why did this use to be a `NamespaceID` and now is bytes? Because now the Graffiti and namespaces differ in size?

liamsi (approved):
LGTM 👍

specs/data_structures.md
adlerjohn: Because we don't necessarily want to change the size of graffiti. It being big enough to put an arbitrary hash or some other data is important I think.

liamsi: That makes sense! I guess I was just wondering why it previously was typed as a `NamespaceID` then.

adlerjohn: Same

Issue #69, opened 2020-09-14
Splitting into shares needs to know request length


adlerjohn:
Currently [shares](https://github.com/lazyledger/lazyledger-specs/blob/7e35c6bbbee59b8e687c94c8483a550d7e68d330/specs/data_structures.md#share) for transactions, evidence, and intermediate state roots are [arranged](https://github.com/lazyledger/lazyledger-specs/blob/7e35c6bbbee59b8e687c94c8483a550d7e68d330/specs/data_structures.md#arranging-available-data-into-shares) by first serializing an array of e.g. transactions, then splitting them into 255-byte shares (with the 256th byte reserved to indicate the starting location of the first start of a transaction in the share.

The issue here is that simply serializing an array of transactions doesn't return the length of each serialized transaction, so we can't properly compute the reserved byte value.

Splitting into shares should instead use the following procedure:
1. Serialize each transaction _individually_ (which tells us the number of bytes for each transaction)
1. Create the byte array `serialize(length of serialize(tx1)) || serialize(tx1) || serialize(length of serialize(tx2)) || serialize(tx2) || ...`
1. Split the above byte array into 255-byte shares, and use the lengths from step (1) to compute reserved byte values

liamsi:
using protobuf this can be achieved without any manual implementation (at least in most languages):
https://developers.google.com/protocol-buffers/docs/techniques#streaming

Otherwise this is simple enough to implement in any language.

liamsi:
Two remarks here:

In 1. the individual serialzation of a TX does it include the lengths or not? Let's say Tx1 is 300 bytes of encoding, then 2. gives us:
`serialize(length of serialize(tx1)) || serialize(tx1) || serialize(length of serialize(tx2)) || serialize(tx2)`
So, If we split it up in 3. do the reserved byte value tell us where `serialize(length of serialize(tx2))` starts, or, where `serialize(tx2)` starts?

adlerjohn:
The reserved byte should tell us where `serialize(length of serialize(tx2))` starts, otherwise we wouldn't know how many bytes to use to decode a tx.

> the computed lengths also include the length delimiter overhead

I don't think that's needed? The length is always a single integer, so a transaction can be parsed in two steps (first parse the length starting at the right offset, then parse the next length bytes as a tx).

liamsi:
OK, I think you are right.

PR #71, opened 2020-09-17
Clarify how shares are split

adlerjohn:
Based on #69, clarify how data is split into shares.
1. Prepend length
1. Clarify serialization and reserved bytes


specs/data_structures.md
```diff
@@ -425,47 +424,51 @@ If a malicious block producer incorrectly computes the 2D Reed-Solomon code for
...
 
 The previous sections described how some original data, arranged into a `k * k` matrix, can be extended into a `2k * 2k` matrix and committed to with NMT roots. This section specifies how [available data](#available-data) (which includes [transactions](#transactiondata), [intermediate state roots](#intermediatestaterootdata), [evidence](#evidencedata), and [messages](#messagedata)) is arranged into the matrix in the first place.
 
- First, for each of `transactionData`, `intermediateStateRootData`, and `evidenceData`, [serialize](#serialization) the data and split it up into `SHARE_SIZE-SHARE_RESERVED_BYTES`-byte [shares](#share). This data has a _reserved_ namespace ID, and as such the first `SHARE_RESERVED_BYTES` bytes for these shares has special meaning. Then, concatenate the lists of shares in the order: transactions, intermediate state roots, evidence. Note that by construction, each share only has a single namespace, and that the list of concatenated shares is [lexicographically ordered by namespace ID](consensus.md#reserved-namespace-ids).
+Then,
+1. For each of `transactionData`, `intermediateStateRootData`, and `evidenceData`, [serialize](#serialization):
```
musalbas, line 464: This seems to be missing messages?

specs/data_structures.md
```diff
@@ -425,47 +424,51 @@ If a malicious block producer incorrectly computes the 2D Reed-Solomon code for
...
+A share is a fixed-size data chunk associated with a namespace ID, whose data will be erasure-coded and committed to in [Namespace Merkle trees](#namespace-merkle-tree).
+
+An example layout of the share's internal bytes is shown below. For shares with a reserved namespace ID through [`NAMESPACE_ID_MAX_RESERVED`](./consensus.md#constants), the first [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes (the `*` in the figure) is the starting byte of the length of the [canonically serialized](#serialization) first request that starts in the share, or `0` if there is none, as a [canonically serialized](#serialization) big-endian unsigned integer. In this example, with a share size of `256` the first byte would be `80` (or `0x50` in hex). For shares with a namespace ID above [`NAMESPACE_ID_MAX_RESERVED`](./consensus.md#constants), the first [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes have no special meaning and are simply used to store data like all the other bytes in the share.
```
musalbas, line 453: It's still ambiguous about whether this share structure is used for all shares or not. I would explicitly mention that shares within the reserved namespace IDs must follow this specific format, and other non-reserved namespace IDs do not, instead of just mentioning the starting byte.

specs/data_structures.md
adlerjohn: Messages are further down, after the figure.

musalbas:
Not part of this commit, but I think this needs to be clarified this is not consensus critical, maybe adding "can optionally".

>However, with only the rule above, interaction between the block producer and transaction sender is required to compute a commitment to the message the transaction sender can sign over. To remove interaction, messages can also be laid out using a non-interactive default:

specs/data_structures.md
adlerjohn: Actually, I do mention this explicitly as

> For shares with a reserved namespace ID through `NAMESPACE_ID_MAX_RESERVED`...

and

> For shares with a namespace ID above `NAMESPACE_ID_MAX_RESERVED`, the first `SHARE_RESERVED_BYTES` bytes have no special meaning and are simply used to store data like all the other bytes in the share.

adlerjohn:
> [Not part of this commit, but I think this needs to be clarified this is not consensus critical, maybe adding "can optionally".](https://github.com/lazyledger/lazyledger-specs/pull/71#issuecomment-699004874)

Fixed in 111b75a.

specs/data_structures.md
liamsi: IMO, describing this part (" the share's internal bytes") in a short pseudo code snippet could be much clearer. Could be added in followup PR though.

liamsi (approved):
👍

musalbas:
The issue is that this description makes it seem like all shares follow 
this specific format, except that the first byte is not used for 
non-reserved namespaces, but all the other parts of the format are still 
used. It's still pretty ambiguous and confusing for a reader unfamiliar 
with our discussions.

PR #75, opened 2020-09-30
Add next validators root to block header

adlerjohn:
Re-add Merkle root for the next block's validators to block header. The current validators root isn't needed, since it's the previous block's next validators root.


musalbas:
I think this needs to be a part of the state commitment because a fraud proof needs to be generated if `nextValidatorsRoot` is invalid.

liamsi:
Can't this be part of the state commitment while still keeping (adding back) that field?

musalbas:
No, because the next validator data in the state commitment might not match `nextValidatorsRoot` in the header.

adlerjohn:
The current validator set is part of the state tree, and is changed after every transaction (and applied at the end of a block). This allows us to have a state transition fraud proof for each _change_ to the validator set.

Now, for proving the Tendermint next validator set root (the subject of this PR): you can simply show that the validator set at the very end of the block doesn't Merkleize to the same root as the Tendermint next validator set root. This is just 100-ish leaves Merkleized twice (once for the state tree and once for the Tendermint header).

One last point of confusion is off-by-one. If the next validator set depends on the current block's transactions, then we need to immediately execute transactions. Thoughts @liamsi?

musalbas:
If the next validator set is already a part of the state tree, why can't nodes simply use the state tree, instead of explicitly adding another commitment for this in the header?

adlerjohn:
> If the next validator set is already a part of the state tree, why can't nodes simply use the state tree, instead of explicitly adding another commitment for this in the header?

From the point of view of LazyLedger, there's no issue with this. From the point of view of the Tendermint _implementation_, it doesn't support the ABCI app doing anything with consensus (e.g. tallying votes). The ABCI app should simply return the new validator set after each block. So...this would be an implementation artifact that ideally we would want to avoid.

liamsi:
> One last point of confusion is off-by-one. If the next validator set depends on the current block's transactions, then we need to immediately execute transactions. Thoughts @liamsi?

The validator hash (in a block of height h) represents the validators that commit this block (and are the output of the prev. block) and the next validators hash will commit the block at height h+1. The tx for the valset changes are always the output from the previous block. Does that cause a problem?

You and others brought up that we might not need both hashes (as e.g. the `ValidatorsHash` of the current block are included as the `NextValidatorsHash` of the previous block). 
Currently, they are used in the tendermint light client e.g. to verify adjacent / consecutive headers (`trustedHeader.NextValidatorsHash == untrustedNewHeader.ValidatorsHash`) and also to validate blocks before execution by checking state's current valset matches the block's valset (`block.ValidatorsHash == state.Validators.Hash()`).

musalbas:
>From the point of view of LazyLedger, there's no issue with this. From the point of view of the Tendermint implementation, it doesn't support the ABCI app doing anything with consensus (e.g. tallying votes). The ABCI app should simply return the new validator set after each block. So...this would be an implementation artifact that ideally we would want to avoid, but in practice might want to have.

Modifying the LazyLedger block validity rule to have two types of data availability fraud proofs instead of one, because of Tendermint implementation limitations, is incredibly hacky and I think we should be trying very hard to avoid this.

liamsi:
Why would that imply "two types of data availability fraud proofs instead of one"?

musalbas:
You would need a different type of fraud proof to prove that the data in the validator root doesn't match the data in the main data root.

liamsi:
OK, makes sense. IMO, that would be a very simple kind of proof though. 
The general tendermint light client relies on both these roots. And some of the minimal state validation needed for tendermint to work also relies on them. Hence, I doubt this is a change will ever make upstreamed back into tendermint (which shouldn't hold us back to modify our fork/ll-core in that way).

Regarding the light-client: it shouldn't be too difficult to remove the checks and have a modified version that verifies the commitments to validator sets in the state (sub)tree instead.
Regarding the validation in the tendermint state package:
https://github.com/lazyledger/lazyledger-core/blob/7b84a4c74317453c6fd4192b1b82619328b1b97c/state/state.go#L48
I think it might be OK to remove [this](https://github.com/lazyledger/lazyledger-core/blob/7b84a4c74317453c6fd4192b1b82619328b1b97c/state/validation.go#L75-L78) as the [state hash](https://github.com/lazyledger/lazyledger-core/blob/7b84a4c74317453c6fd4192b1b82619328b1b97c/state/validation.go#L56-L61) is also verified to match the block's state hash and the available data root would be checked for validity too.

> How hard would it be to modify Tendermint so that this is not the case?

Not impossible. It would further blur the boundary between the app and tendermint a bit. I think I'll start a branch to answer this more precisely.

liamsi:
Do the same concerns hold for the other roots? E.g. looking at the current version of the header, the consensus root? https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#header

musalbas:
Aren't the consensus parameters hardcoded? I don't see why they even need to be committed to in the block header.

liamsi:
So this branch removes all traces from `ValidatorsHash` and `NextValidatorsHash`: https://github.com/lazyledger/lazyledger-core/compare/ismail/unsafe_removal_ofvalhashes?expand=1
Of course, we can't just remove the checks in the light client code like this! So making this all safe is the actual work behind this. Some of that work will be inside of ll-core (data availability checks, the light client checks) and some of it will happen in the (abci) lazyledger app (tracking the validator sets in the state tree and related checks).

> Aren't the consensus parameters hardcoded? I don't see why they even need to be committed to in the block header.

In tendermint Consensus Parameters can change with every block:
https://docs.tendermint.com/master/spec/abci/apps.html#consensus-parameters

I see that the consensus params that made it into our specs are only very few and most of them won't change in the lifetime of a chain/without a fork anyways I guess? 
Only `availableDataOriginalSquareSize` is something that is likely to change for each block as it depends on the number of tx in the block (cc @adlerjohn).

adlerjohn:
> [Aren't the consensus parameters hardcoded? I don't see why they even need to be committed to in the block header.](https://github.com/lazyledger/lazyledger-specs/pull/75#issuecomment-703913970)

Essentially, yes. Those parameters should only change when there's a hard fork. **Note that this part of the spec details data structures as they are signed over, _not_ data structures as they are stored (an implementation detail), or sent over the network.** Common and fixed values can be optimized away at the network layer (e.g. a node indicates which consensus root it communicates with, then all subsequent blocks received by that node implicitly use the same consensus root).

> [Only `availableDataOriginalSquareSize` is something that is likely to change for each block as it depends on the number of tx in the block](https://github.com/lazyledger/lazyledger-specs/pull/75#issuecomment-703924868)

That should probably be renamed to `availableDataOriginalSquareSizeMax`, to indicate the largest blocksize accepted by the network, rather than the size of the current block (which should be placed elsewhere in the block header).

liamsi:
> Essentially, yes. Those parameters should only change when there's a hard fork.

Then these are more like constants which are different from the consensus params in tendermint (which could essentially change every block).

> That should probably be renamed to availableDataOriginalSquareSizeMax

Let's track this in a separate issue!

liamsi:
closing in favour of removing the validator roots altogether: #78

Issue #23, opened 2020-05-06
Initial evidence types


adlerjohn:
Currently the only evidence type is double-signing. There should be additional types for anything slashable, i.e. anything that would trigger a fraud proof.

liamsi:
Additionally to defining the exact data structures of different fraud proofs / evidence types, the specification should also stipulate who (which node types) generates them under which circumstances and what happens with the evidence after construction (does get included in the next block, does it trigger a slashing event, is it "only" gossipped in the p2p network - if yes to which nodes). This to actually guide the implementation.

In the context of erasure coding, for instance, this is needed to inform the APIs of the used libraries (e.g. see https://github.com/lazyledger/rsmt2d/pull/24 and github.com/lazyledger/rsmt2d#12).  

Ideally, the spec also defined properties or invariants (if any) that are achieved by dispersing the evidence (e.g. a proposer can not propose an incorrectly encoded block without getting caught and getting slashed for that, etc).

PR #76, opened 2020-10-01
update duplicatevote_evidence

marbar3778:
Update duplicateVote Evidence with the repo. 

Would you want to add the other types of evidence currently supported in tendermint/ll-core?

Sorry for the lint fixes.

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -18,6 +18,7 @@ Data Structures
 - [Serialization](#serialization)
 - [Hashing](#hashing)
 - [Public-Key Cryptography](#public-key-cryptography)
+  - [PublicKey](#publickey)
 - [Merkle Trees](#merkle-trees)
   - [Binary Merkle Tree](#binary-merkle-tree)
     - [Binary Merkle Tree Proofs](#binary-merkle-tree-proofs)
@@ -53,7 +54,6 @@ Data Structures
     - [IntermediateStateRoot](#intermediatestateroot)
   - [EvidenceData](#evidencedata)
     - [Evidence](#evidence)
-    - [PublicKey](#publickey)
     - [Vote](#vote)
   - [MessageData](#messagedata)
     - [Message](#message)
@@ -166,13 +166,6 @@ Address is a [type alias](#type-aliases).
 
 Addresses are the [hash](#hashing) [digest](#hashdigest) of the [public key](#publickey).
 
-### PublicKey
-
-| name | type       | description              |
-| ---- | ---------- | ------------------------ |
-| `x`  | `byte[32]` | `x` value of public key. |
-| `y`  | `byte[32]` | `y` value of public key. |
-
 ### CommitSig
 
 ```C++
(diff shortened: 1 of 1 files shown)


specs/data_structures.md
```diff
@@ -166,6 +166,13 @@ Address is a [type alias](#type-aliases).
 
 Addresses are the [hash](#hashing) [digest](#hashdigest) of the [public key](#publickey).
 
+### PublicKey
```
adlerjohn, line 169: Should fix the ToC order too

adlerjohn: Actually, `PublicKey` seems to only have been used for evidence. Is there a need for it? Why remove `PublicKey` from equivocations? Because it can be recovered from the signature?

adlerjohn:
> Would you want to add the other types of evidence currently supported in tendermint/ll-core?

Wouldn't hurt. #23

specs/data_structures.md
marbar3778: The pubkey was used to get the address of the validator that committed the infraction but the votes contained the same info, so it was redundant. 

I can remove it entirely

marbar3778: actually I think it may be better to key the pubkey becuase it is referenced elsewhere

adlerjohn: Ah right, votes have both an address and signature (so technically the address can be removed from there too, but would involve non-trivial changes to Tendermint, so we can revisit it later). The pubkey isn't used anywhere else, so probably best to just remove it.

specs/data_structures.md
```diff
@@ -232,6 +225,13 @@ Putting it all together, the encoding for signatures is:
 
 This encoding scheme is derived from [EIP 2098: Compact Signature Representation](https://eips.ethereum.org/EIPS/eip-2098).
 
+### PublicKey
+
+| name | type       | description              |
+| ---- | ---------- | ------------------------ |
+| `x`  | `byte[]` | `x` value of public key. |
+| `y`  | `byte[]` | `y` value of public key. |
```
adlerjohn, line 233: Fixed-size arrays should have their size in the spec (i.e. `byte[32]` here). Sure protobuf doesn't have fixed-size arrays, but other serialization schemes might, and we don't want the specs to depend on the internals or restrictions of the serialization scheme.

specs/data_structures.md
```diff
@@ -723,11 +723,55 @@ Wrapper for evidence data.
...
+| `proposer`  | [Validator](#validator)           |             |
+
+<!-- #### Validator
+
+| name     | type                    | description |
+| -------- | ----------------------- | ----------- |
+| `address`  | [Address](#address)           |             |
+| `pubKey`  | [publickey](#publickey)           |             |
+| `VotingPower`  | int64           |             |
+| `ProposerPriority`  | int64           |             | -->
```
marbar3778, line 773: @adlerjohn within this page you have a validator type, but the type you have defined differs from this one and the one that will be used within ll-core. How would you prefer to handle this?

This may tie into the conversation on how to define ll-core specific types.

liamsi:
What's with this PR? Is the plan to merge it in eventually? If not, let's close it instead.

marbar3778:
never got an answer to the question, left it as is. It updates the evidence types in order to better describe whats in ll-core. I guess we can close it and push this till al later date.

liamsi:
cc @adlerjohn

adlerjohn:
I think we can close this for now until we loop back to evidence handling (which might be in a little while since evidence isn't super-critical for devnet).

PR #79, opened 2020-10-07
Variable available data size

adlerjohn:
- Add a new field in the header for the available data size for the current block.
- Rename the consensus parameter to max size.

Diff without whitespace changes: https://github.com/lazyledger/lazyledger-specs/pull/79/files?diff=unified&w=1


specs/figures/block_data_structures.dot
```diff
@@ -22,12 +22,12 @@ digraph G {
 
         subgraph cluster_header {
             label = "header";
-            struct1 [label = "version | chainID | height | time | lastBlockID | <f5> lastCommitRoot | consensusRoot | stateCommitment | <f8> availableDataRoot | proposerAddress"];
+            struct1 [label = "height | timestamp | lastBlockID | <f3> lastCommitRoot | consensusRoot | stateCommitment | <f6> availableDataRoot | proposerAddress"];
```
liamsi, line 25: Not sure it is a good idea to remove the chain ID from the header.

adlerjohn: The chainID and version is in the consensus parameters. Since those parameters are fixed and unchangeable without a hard fork (I know Tendermint allows changing parameters with approval from validators, but that's something we want to avoid), there's not much point in duplicating all these values in the top-level header when they can instead just be committed to once.

liamsi: Reason is that the chainID in the header is used to add it to the votes. Maybe instead we can hardcode it into our code base instead (and not have it in the header). Not sure. It is necessary in the votes at least. Otherwise a LL validator that uses the same key to sign multiple other chains could e.g. be slashed for that.

liamsi: Just saw your comment. As long as we commit to the value once and it is accessible to sign votes, then that is fine.

specs/data_structures.md
```diff
@@ -105,16 +105,17 @@ Blocks are the top-level data structure of the LazyLedger blockchain.
...
+| `lastCommitRoot`                  | [HashDigest](#hashdigest) | Previous block's Tendermint commit root.                                                                                                                                              |
+| `consensusRoot`                   | [HashDigest](#hashdigest) | Merkle root of [consensus parameters](#consensus-parameters) for this block.                                                                                                          |
+| `stateCommitment`                 | [HashDigest](#hashdigest) | The [state root](#state) after this block's transactions are applied.                                                                                                                 |
+| `availableDataOriginalSquareSize` | `uint64`                  | The number of rows/columns of the original data [shares](data_structures.md#share) in [square layout](data_structures.md#arranging-available-data-into-shares). Must be a power of 2. |
```
liamsi, line 116: ~~Must be a power of 2.~~
Must be a square number (?)

specs/data_structures.md
```diff
@@ -385,9 +386,9 @@ Note that while data is laid out in a two-dimensional square, rows and columns a
 
 Reed-Solomon erasure coding is used as the underlying coding scheme. The parameters are:
 - 16-bit Galois field
-- `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE` original pieces
-- `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE` parity pieces (i.e `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE * 2` total pieces), for an erasure efficiency of 50%. In other words, any 50% of the pieces from the `AVAILABLE_DATA_ORIGINAL_SQUARE_SIZE * 2` total pieces are enough to recover the original data.
-- `SHARE_SIZE` bytes per piece
+- [`availableDataOriginalSquareSize`](#header) original pieces (maximum of [`AVAILABLE_DATA_ORIGINAL_SQUARE_MAX`](./consensus.md#constants))
```
liamsi, line 389: is it the number of shares or the number rows/cols as mentioned above?

specs/data_structures.md
adlerjohn: It's the number of rows/columns, not rows times columns. (Philosophically, you can define a rectangle as the length of two of its sides, and a square as the length of one of its sides since a square is a special rectangle).

Technically there's no absolutely hard reason we would need to restrict the size of the original square to a power of 2, but it'll make our life easier and there's not any particularly meaningful downside that I can see.

specs/data_structures.md
adlerjohn: Both? If there are X rows, then a column has X shares by construction.

liamsi: If a column has X shares, shouldn't the total number of original pieces (shares) should be X times X and not only X?

adlerjohn: Each erasure coding operation happens on a single row or column, not on the entire square.

specs/data_structures.md
liamsi: In what way does a power of 2 make our life easier?

specs/data_structures.md
liamsi: OK, so this doesn't refer to the total number of original pieces but the number of original pieces per row (or column). I think this becomes clearer where we talk about the 2D scheme.

specs/data_structures.md
adlerjohn: If the number of shares is a power of 2, then we have the guarantee that the entire left subtree of the NMT is original data and the entire right subtree is parity data.

liamsi: While I see the beauty of this is it really worth blowing up the square? Are there other reasons why this is desireable?

adlerjohn: It also gives us the nice property that we can compute the structures of the paths of subtrees of different sizes and starting offsets trivially, which is needed for the non-interactive default message layout.

musalbas: An alternative would be to define a wrapper function for verifying Merkle proofs via the data root that offsets the index based on the size of each row and column, see section 5.2 in https://arxiv.org/pdf/1809.09044.pdf and the footnote.

adlerjohn: Ah, so here's the issue with not using a power of 2: consider the case of a max square size of 8x8. A user wants to include a message 8 shares long, and includes a signature in their transaction for a single Merkle root with 8 leaves. Now if the block producer makes a 7x7 block, the transaction won't be valid for that block, since that message would need two rows.

If square sizes are always a power of 2, then transaction senders can non-interactively include a logarithmic number of signatures with their transaction, for different square sizes. Note that the signatures for square sizes > the number of message shares will be duplicates (and can thus be deduplicated).

liamsi: > then transaction senders can non-interactively include a logarithmic number of signatures with their transaction,

I'm not sure I understand how this would work. I think this needs further clarification. Let's say user want to submit Message `(myOruNID, myOruBlob)` to LL. 
What steps does a user need to do before he can submit that message and a Tx that pays for that message? What needs to be generated (the Tx obviously) and what are the things that need to be signed?

liamsi: Also, I think we need to be clearer on how does the block producer decide on the number `k`/`availableDataOriginalSquareSize `.

liamsi:
After our call this became much clearer! 🙏 

I think what needs to be done for this PR to get merged: now that the square size it isn't fixed anymore but depends on the data, the resulting number `availableDataOriginalSquareSize` is a function of the size of Tx and Messages or rather namespaced Shares from both. IMO this should be made clearer here.

In followup PRs we should further clarify:

 - the user-facing side (how many sigs would a user need to create and why); basically the whole process of a Tx creation and signing needs to be in the spec too
 - why a tx per Message is required (and potential alternatives)
 - solidify the idea that a certain portion of the block can be "for free"

Anything else? (cc @musalbas)

specs/data_structures.md
adlerjohn: Scratch [what I said above](https://github.com/lazyledger/lazyledger-specs/pull/79#discussion_r502198374). You should be able to still sign over a logarithmic number of different share layouts regardless of the specific current square size. That being said, the formulas for the NMT currently assume that the square is a power of 2, so if we want to allow more flexibility I can look into it in a future PR that would have to change those too.

adlerjohn:
> [resulting number `availableDataOriginalSquareSize` is a function of the size of Tx and Messages or rather namespaced Shares from both. IMO this should be made clearer here](https://github.com/lazyledger/lazyledger-specs/pull/79#issuecomment-706302260)

Clarified that `availableDataOriginalSquareSize` may vary each block in d1eb1a0. Any additional clarification should probably be in the [block_proposer.md](https://github.com/lazyledger/lazyledger-specs/blob/187d757b577506c4f13c950747c98b3f9cb1f0d6/specs/block_proposer.md) file, which should include some rationale around how block proposers should construct blocks.

specs/data_structures.md
```diff
@@ -390,7 +390,7 @@ Reed-Solomon erasure coding is used as the underlying coding scheme. The paramet
 - [`availableDataOriginalSquareSize`](#header) parity pieces (maximum of [`AVAILABLE_DATA_ORIGINAL_SQUARE_MAX`](./consensus.md#constants)) (i.e `availableDataOriginalSquareSize * 2` total pieces), for an erasure efficiency of 50%. In other words, any 50% of the pieces from the `availableDataOriginalSquareSize * 2` total pieces are enough to recover the original data.
 - [`SHARE_SIZE`](./consensus.md#constants) bytes per piece
 
-[Leopard-RS](https://github.com/catid/leopard) is a C library that implements the above scheme with quasilinear runtime.
+Note that [`availableDataOriginalSquareSize`](#header) may vary each block. [Leopard-RS](https://github.com/catid/leopard) is a C library that implements the above scheme with quasilinear runtime.
```
liamsi, line 393: Can we clarify further how it exactly it varies and what this depends on? Ideally, this could be expressed as a function? Does it depend on the largest Message (which means more shares)?

liamsi:
> Any additional clarification should probably be in the block_proposer.md file, which should include some rationale around how block proposers should construct blocks.

That is OK but it seems like this PR currently only state, that the square sized is not fixed but rather depends on something without saying what that something is.

specs/data_structures.md
adlerjohn: Fixed in 53e5f35.

liamsi (approved):
LGTM now!

Issue #22, opened 2020-05-04
Define different node types


adlerjohn:
There are several different types of nodes supported by LazyLedger, each with different compute and networking requirements, along with different assumptions and security properties. These should be specified.

PR #89, opened 2020-10-27
Initial consensus rules

### LINKED ISSUE #86: Padding shares should be deterministic and consensus-enforced
[Currently](https://github.com/lazyledger/lazyledger-specs/blob/301b88ecfeb0953a97e4f349729deded01a5364f/specs/data_structures.md#arranging-available-data-into-shares), padding between messages can have any namespace ID between the namespace IDs of the surrounding messages:
> The non-interactive default rules may introduce empty shares that do not belong to any message (in the example above, the top-right share is empty). These must be explicitly disclaimed by the block producer using special transactions.

However, these padding shares aren't included when distributing the entire block all at once, since blocks aren't distributed normally laid out as shares. To fix this, the namespace ID of padding shares must be both deterministic and enforced by consensus.

adlerjohn:
Initial pass of the consensus rules, covering block structure and state transitions.

- [x] Block structure, i.e. making sure the format of a block is correct.
- [x] State transitions, i.e. applying transactions.
- [x] Filed follow-up issues: celestiaorg/celestia-app#739 celestiaorg/celestia-specs#92 celestiaorg/celestia-specs#93 celestiaorg/celestia-specs#94 celestiaorg/celestia-specs#95 celestiaorg/celestia-specs#96 celestiaorg/celestia-specs#98.


specs/consensus.md
```diff
@@ -140,10 +142,19 @@ TODO define fraud proof for invalid block here
 
 Once parsed, the following checks must be `true`:
 
-1. The commitments of the [erasure-coded extended](./data_structures.md#2d-reed-solomon-encoding-scheme) `availableData` must match those in `header.availableDataHeader`.
+1. The commitments of the [erasure-coded extended](./data_structures.md#2d-reed-solomon-encoding-scheme) `availableData` must match those in `header.availableDataHeader`. Implicitly, this means that both rows and columns must be ordered lexicographically by namespace ID since they are committed to in a [Namespace Merkle Tree](data_structures.md#namespace-merkle-tree).
+1. Length of `availableData.intermediateStateRootData` == length of `availableData.transactionData` + length of `availableData.evidenceData`.
```
liamsi, line 146: Why would evidence add to the intermediate state roots?

liamsi: Ah OK, that's assuming the evidence has an immediate effect on the state root.

adlerjohn: Yep, which it should! Each evidence actually mutates the state, i.e. is a state transition, and so should have an associated intermediate state root.

rationale/message_block_layout.md
```diff
@@ -46,4 +46,9 @@ The last piece of the puzzle is determining _which_ row the message is placed at
 
 ### Caveats
 
-The message placement rules described above conflict with the first rule that shares must be ordered by namespace ID, as shares between two messages that are not placed adjacent to each other do not have a natural namespace they belong to. This is resolved by having special transactions the block producer includes that specify a range a zero-padding shares for a given namespace ID (which must be between the namespace IDs of the surrounding real shares, inclusive).
+The message placement rules described above conflict with the first rule that shares must be ordered by namespace ID, as shares between two messages that are not placed adjacent to each other do not have a natural namespace they belong to. This is resolved with some rules:
```
musalbas, line 49: Maybe say "padding shares between two messages" instead of "shares"?

rationale/message_block_layout.md
```diff
@@ -46,4 +46,9 @@ The last piece of the puzzle is determining _which_ row the message is placed at
...
+1. the namespace ID for message shares must be "even" (i.e. the lowest bit must be `0`), and
+1. the namespace ID for shares between messages must be exactly `1` more than the namespace ID of the preceding message (i.e. must be identical save for the lowest bit being `1` instead of `0`), and.
```
musalbas, line 51: Likewise here. It doesn't seem to make sense to say "shares between messages", as messages are shares.

rationale/message_block_layout.md
```diff
@@ -46,4 +46,9 @@ The last piece of the puzzle is determining _which_ row the message is placed at
...
+1. the namespace ID for shares between requests with a reserved namespace ID and messages must be [`TAIL_TRANSACTION_PADDING_NAMESPACE_ID`](../specs/consensus.md#constants). Note that this only applies for shares _between_ requests with a reserved namespace ID and messages, i.e. if there are no messages in the block then this rule does not apply.
```
musalbas, line 52: Why do you need a special rule for reserved namespace IDs? Also, what is a "request" in the context of a block?

rationale/message_block_layout.md
```diff
@@ -46,4 +46,9 @@ The last piece of the puzzle is determining _which_ row the message is placed at
...
+
+If we interpret namespace IDs as unsigned integers, this can be thought of as restricting the namespace IDs of shares belonging to messages to even namespace IDs and padding shares to odd namespace IDs, further restricted to deterministic values.
```
musalbas, line 54: I don't think an "if" should be part of a spec. Are namespace IDs unsigned integers or not?

specs/consensus.md
```diff
@@ -71,16 +85,149 @@ Consensus Rules
...
+## Block Structure
+
+Before executing [state transitions](#state-transitions), the structure of the [block](./data_structures.md#block) must be verified.
+
+The following block fields are acquired from the network and parsed (i.e. [deserialized](./data_structures.md#serialization)). If they cannot be parsed, the block is ignored but is not explicitly considered invalid by consensus rules. Further implications of ignoring a block are found in the [networking spec](./networking.md).
+1. [block.header](./data_structures.md#header)
+1. [block.availableDataHeader](./data_structures.md#availabledataheader)
+1. [block.lastCommit](./data_structures.md#commit)
+
+If the above fields are parsed successfully, the available data `block.availableData` is acquired in erasure-coded form as [a list of share rows](./networking.md#availabledata), then parsed. If it cannot be parsed, the block is ignored but not explicitly invalid, as above.
```
musalbas, line 109: The whole data shouldn't need to be acquired due to DA proofs etc.

specs/consensus.md
```diff
@@ -71,16 +85,149 @@ Consensus Rules
...
-A transaction `tx` that requests a new validator initializes a new [Validator](data_structures.md#validator) leaf in the inactive validators subtree for that account as follows:
+## State Transitions
+
+Once the basic structure of the block [has been validated](#block-structure), state transitions must be applied to compute the new state and state root.
+
+For this section, the variable `state` represents the [state tree](./data_structures.md#state), with `state.accounts[k]`, `state.inactiveValidatorSet[k]`, and `state.activeValidatorSet[k]` being shorthand for the leaf in the state tree in the [accounts, inactive validator set, and active validator set subtrees](./data_structures.md#state) with [pre-hashed key](./data_structures.md#state) `k`. E.g. `state.accounts[a]` is shorthand for `state[(ACCOUNTS_SUBTREE_ID << 8*(32-STATE_SUBTREE_RESERVED_BYTES)) | ((-1 >> 8*STATE_SUBTREE_RESERVED_BYTES) & hash(a))]`.
+
+### `block.availableData.evidenceData`
+
+Evidence is the first set of state transitions that are applied, and represent proof of validator misbehavior.
```
musalbas, line 163: Don't you mean the first set that represent proof of validator misbehaviour? Can you expand on what this evidence is for, and why it needs to be committed in the header?

specs/consensus.md
```diff
@@ -71,16 +85,149 @@ Consensus Rules
...
+```
+
+#### SignedTransactionDataCreateValidator
+
+The following checks must be `true`:
+
+1. `tx.type` == [`TransactionType.CreateValidator`](./data_structures.md#signedtransactiondata).
+1. `tx.amount` <= `state.accounts[sender].balance`.
+1. `tx.nonce` == `state.accounts[sender].nonce + 1`.
+1. `tx.commissionRate` <!--TODO check some bounds here-->
```
musalbas, line 220: Add todo as issues

musalbas (dismissed):

specs/consensus.md
liamsi: I think John already opened issues for them, e.g.: https://github.com/lazyledger/lazyledger-specs/issues/95
These could be a bit more descriptive IMO but I guess that's OK.

specs/consensus.md
adlerjohn: For now the assumption is a full node fully downloads the entire block, so as to check everything (including that the erasure coding was done correctly). Later on we can specifically notate which functionality is required for each node type, as we define node types in #22.

specs/consensus.md
adlerjohn: I don't think so? I mean that they're the first set of state transitions that are applied. I.e. each evidence is a state transition, and evidence is applied before transactions. I'd prefer to relegate full explanation of what evidence is to a future issues, e.g. #23.

rationale/message_block_layout.md
liamsi: > Also, what is a "request" in the context of a block?

I also found that word confusing:
https://github.com/lazyledger/lazyledger-specs/pull/21#discussion_r419653996

Maybe we should instead just say Tx or rather "shares with a reserved namespace ID" instead of requests?

specs/consensus.md
adlerjohn: Yeah, the rest of the TODOs were removed in 1e2df94, except for this one so that I don't forget since it's in the middle of a bunch of other text as opposed to a discrete concept.

rationale/message_block_layout.md
adlerjohn: Well, I use "request" because transactions aren't the only things that touch state: https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r529003291

adlerjohn: > Why do you need a special rule for reserved namespace IDs?

Actually, I realized there's an issue here: if two messages both have the same namespace ID, then they can't both be included in a block under this scheme unless they are aligned to some nice power of 2. It might have been a better idea to simply have a special transaction type that pays for padding, and revert the changes made in this PR to target #86.

> Maybe we should instead just say Tx or rather "shares with a reserved namespace ID" instead of requests?

"shares with a reserved namespace ID" works here.

specs/consensus.md
musalbas: I see, I thought it meant that the evidence itself is the first set of transactions that were applied, rather that it meaning that the evidence should be applied first. Is evidence really a state transition to be applied?

adlerjohn: Right, evidence causes slashing, which is a state transition. Wording could be improved on that. More context: https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r521342908

rationale/message_block_layout.md
adlerjohn: Reverted in 6d9648a as per https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r529009826.

rationale/message_block_layout.md
adlerjohn: Reverted in 6d9648a as per https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r529009826.

rationale/message_block_layout.md
adlerjohn: Reverted in 6d9648a as per https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r529009826.

rationale/message_block_layout.md
adlerjohn: Reverted in 6d9648a as per https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r529009826.

specs/consensus.md
adlerjohn: Clarified wording in 37d31ca.

adlerjohn:
Since the last review, the major change has been reverting the change that were previously made for celestiaorg/celestia-specs#86 in 6d9648a as per https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r529009826, and adding the corresponding state transition rules for the transaction that pays for padding.

rationale/message_block_layout.md
```diff
@@ -46,4 +46,4 @@ The last piece of the puzzle is determining _which_ row the message is placed at
 
 ### Caveats
 
-The message placement rules described above conflict with the first rule that shares must be ordered by namespace ID, as shares between two messages that are not placed adjacent to each other do not have a natural namespace they belong to. This is resolved by having special transactions the block producer includes that specify a range a zero-padding shares for a given namespace ID (which must be between the namespace IDs of the surrounding real shares, inclusive).
+The message placement rules described above conflict with the first rule that shares must be ordered by namespace ID, as shares between two messages that are not placed adjacent to each other do not have a natural namespace they belong to. This is resolved by having [special transactions](./../specs/data_structures.md#signedtransactiondatapayforpadding) the block producer includes that specify a range a zero-padding shares for a given namespace ID (which must be between the namespace IDs of the surrounding real shares, inclusive).
```
liamsi, line 49: Does this mean that it becomes required to send this additional Tx? So if I want to include a message, I'll have to send a Tx that pays for a message and for padding?

musalbas: If that's the case, I think this introduces some undesirable complexity.

liamsi: also there seems to be a typo in 
> the block producer includes that specify a range a zero-padding shares 

How does the user know what range to pay for?

adlerjohn: These are created by the block proposer themselves, i.e. not broadcast through the p2p network. So there's no typo.

specs/consensus.md
```diff
@@ -30,28 +45,29 @@ Consensus Rules
...
 | `AVAILABLE_DATA_ORIGINAL_SQUARE_MAX` | `uint64` |         | `share` | Maximum number of rows/columns of the original data [shares](data_structures.md#share) in [square layout](data_structures.md#arranging-available-data-into-shares). |
+| `CHAIN_ID`                           | `uint64` | `1`     |         | Chain ID. Each chain assigns itself a (unique) ID.                                                                                                                  |
 | `GENESIS_COIN_COUNT`                 | `uint64` | `10**8` | `4u`    | `(= 100000000)` Number of coins at genesis.                                                                                                                         |
-| `UNBONDING_DURATION`                 | `uint32` |         | `block` | Duration, in blocks, for unbonding a validator or delegation.                                                                                                       |
+| `GRAFFITI_BYTES`                     | `uint64` | `32`    | `byte`  | Maximum size of transaction graffiti, in bytes.                                                                                                                     |
```
liamsi, line 51: Seem a bit more consistent with the other max:
```suggestion
| `MAX_GRAFFITI_BYTES`                     | `uint64` | `32`    | `byte`  | Maximum size of transaction graffiti, in bytes.                                                                                                                     |
```

specs/consensus.md
```diff
@@ -30,28 +45,29 @@ Consensus Rules
...
 | `MAX_VALIDATORS`                     | `uint16` | `64`    |         | Maximum number of active validators.                                                                                                                                |
+| `NAMESPACE_ID_BYTES`                 | `uint64` | `8`     | `byte`  | Size of namespace ID, in bytes.                                                                                                                                     |
+| `NAMESPACE_ID_MAX_RESERVED`          | `uint64` | `254`   |         | Value of maximum reserved namespace ID (inclusive). 1 byte worth of IDs.                                                                                            |
```
liamsi, line 54: Does this correspond to `0xFFFFFFFFFFFFFFFF`?

rationale/message_block_layout.md
musalbas: I don't understand. The description for https://github.com/lazyledger/lazyledger-specs/pull/89/commits/6d9648a454bcbfef79172152a4f129253cb7962d#diff-6aed1ad746ae3318f74b4ee4094dda4664865e75bf4b8a3c95e58773292d1c74R603 says that this special transaction pays for padding. Who's paying for the padding if not the user, and how is it that block producers can generate these transactions themselves?

specs/consensus.md
adlerjohn: Oh, I forgot the revert that. It should be 255, which corresponds to `0xFF`.

rationale/message_block_layout.md
liamsi: The block proposer pays for padding?

>So there's no typo.
Shouldn't that be: that specify a range *of* zero-padding shares?

specs/consensus.md
```diff
@@ -71,16 +86,160 @@ Consensus Rules
...
+The following block fields are acquired from the network and parsed (i.e. [deserialized](./data_structures.md#serialization)). If they cannot be parsed, the block is ignored but is not explicitly considered invalid by consensus rules. Further implications of ignoring a block are found in the [networking spec](./networking.md).
+1. [block.header](./data_structures.md#header)
+1. [block.availableDataHeader](./data_structures.md#availabledataheader)
+1. [block.lastCommit](./data_structures.md#commit)
+
+If the above fields are parsed successfully, the available data `block.availableData` is acquired in erasure-coded form as [a list of share rows](./networking.md#availabledata), then parsed. If it cannot be parsed, the block is ignored but not explicitly invalid, as above.
+
+### `block.header`
+
+The [block header](./data_structures.md#header) `block.header` (`header` for short) is the first thing that is downloaded from the new block, and commits to everything inside the block in some way. For previous block `prev` (if `prev` is not known, then the block is ignored), and previous block header `prev.header`, the following checks must be `true`:
```
liamsi, line 114: > The [block header](./data_structures.md#header) `block.header` (`header` for short) is the first thing that is downloaded from the new block

If it is actively downloaded or received (e.g. via some gossip routine etc) should go in the network spec. Maybe say "processed", this implies that you have it at hand.

rationale/message_block_layout.md
adlerjohn: > Who's paying for the padding if not the user

The block proposer. They know what's in the block so they can make new txs. Note that I haven't actually decided if a fee should be paid for these, and if not then there's not much point in having them as txs vs just part of the block data that notate inter-message padding. That's for a future issue.

liamsi: I find it weird that the block proposer would pay for these shares. Why wouldn't they simply drop all the relevant messages instead.

musalbas: Yeah, I'm not sure about the economic rationale behind that.

adlerjohn: > Yeah, I'm not sure about the economic rationale behind that.

Agree that it's not clear either way, hence me saying that not having this information as transactions is potentially the way to go, but it'll require further investigation that's outside the scope of this PR.

adlerjohn: Filed #98 to track this issue.

specs/consensus.md
adlerjohn: Fixed in 9a31d32.

specs/consensus.md
adlerjohn: The plan is actually to move that entire paragraph to the networking section, since a node won't even accept a block header unless it passes those checks.

specs/consensus.md
```diff
@@ -71,16 +86,160 @@ Consensus Rules
...
+#### SignedTransactionDataPayForPadding
+
+1. `tx.type` == [`TransactionType.PayForPadding`](./data_structures.md#signedtransactiondata).
+1. The `ceil(tx.messageSize / SHARE_SIZE)` shares starting at index `wrappedTransactions.messageStartIndex` must:
+    1. Have namespace ID `tx.messageNamespaceID`.
+
+Apply the following to the state:
+
+```
+```
```
liamsi, line 223: Shouldn't we remove this type together with the paragraph under Caveats?
(https://github.com/lazyledger/lazyledger-specs/pull/89/files#r531078625)

adlerjohn: I don't want to be overzealous with the scope of this PR, so I filed #98 to follow up on how padding should be handled. For now I'd rather just leave the PayForPadding tx type as it was. https://github.com/lazyledger/lazyledger-specs/pull/89#discussion_r531129652

liamsi (approved):

PR #90, opened 2020-11-04
Add field for number of shares used, split row and column roots

adlerjohn:
Add a field in the header for number of shares used that aren't tail padding, and describe how to compute the number of row and column roots. Also prune rows that only contain tail padding.

The splitting up of roots into rows and columns doesn't really affect anything since the arrays aren't directly part of the block header, but are instead committed to in the block header.

Diff without whitespace changes: https://github.com/lazyledger/lazyledger-specs/pull/90/files?diff=unified&w=1


specs/data_structures.md
```diff
@@ -105,23 +105,29 @@ Blocks are the top-level data structure of the LazyLedger blockchain.
...
-| -------------------------- | ----------------------------- | -------------------------------------- |
-| `availableDataCommitments` | [HashDigest](#hashdigest)`[]` | Commitments to all erasure-coded data. |
+| name       | type                          | description                            |
+| ---------- | ----------------------------- | -------------------------------------- |
+| `rowRoots` | [HashDigest](#hashdigest)`[]` | Commitments to all erasure-coded data. |
+| `colRoots` | [HashDigest](#hashdigest)`[]` | Commitments to all erasure-coded data. |
+
+The `availableDataRoot` of the [header](#header) is computed using the row and column roots as described [here](#2d-reed-solomon-encoding-scheme).
+
+The number of column roots is twice the [header](#header)'s `availableDataOriginalSquareSize` field, and the number of row roots is `ceil(availableDataOriginalSharesUsed / availableDataOriginalSquareSize) + availableDataOriginalSquareSize`.
```
liamsi, line 130: So #column roots = 2 * availableDataOriginalSquareSize because it contains all the parity shares, too. And for the row roots, we only compute the roots that do not only contain tail padding. Maybe add the latter as a sentence before or after this?

specs/data_structures.md
```diff
@@ -413,11 +419,11 @@ As an example, the parity data in the second column of `Q2` (in striped purple)
 
 ![fig: RS2D encoding: extending a column.](./figures/rs2d_extend.svg)
 
-Now that all four quadrants of the `2k * 2k` matrix are filled, the row and column roots can be computed. To do so, each row/column is used as the leaves of a [NMT](#namespace-merkle-tree), for which the compact root is computed (i.e. an extra hash operation is used to produce a single [HashDigest](#hashdigest)). In this example, the fourth row root value is computed as the NMT root of the fourth row of `Q0` and the fourth row of `Q1` as leaves.
+Now that all four quadrants of the `2k * 2k` matrix are filled, the row and column roots can be computed. To do so, each row/column is used as the leaves of a [NMT](#namespace-merkle-tree), for which the compact root is computed (i.e. an extra hash operation over the NMT root is used to produce a single [HashDigest](#hashdigest)). In this example, the fourth row root value is computed as the NMT root of the fourth row of `Q0` and the fourth row of `Q1` as leaves.
```
liamsi, line 422: Question: Is the compact root a hard requirement here? Especially given that a single and compact commitment to all roots will end up as `availableDataRoot` in the header.

liamsi (approved):
An optional nit about making it more accessible why #rows != #cols. Otherwise LGTM 👍

specs/data_structures.md
adlerjohn: It's not a hard requirement, for data availability proofs, however it does reduce the size of the commitments that need to be downloaded (which over time adds up). Nodes have to download some commitment to each NMT they sample into, so might as well make that as small as possible.

specs/data_structures.md
adlerjohn: Fixed in 64a08e7.

liamsi (dismissed):

specs/data_structures.md
```diff
@@ -105,23 +105,29 @@ Blocks are the top-level data structure of the LazyLedger blockchain.
...
+| `consensusRoot`                   | [HashDigest](#hashdigest) | Merkle root of [consensus parameters](#consensus-parameters) for this block.                                                                                                       |
+| `stateCommitment`                 | [HashDigest](#hashdigest) | The [state root](#state) after this block's transactions are applied.                                                                                                              |
+| `availableDataOriginalSharesUsed` | `uint64`                  | The number of shares used in the [original data square](#arranging-available-data-into-shares) that are not [tail padding](./consensus.md#reserved-namespace-ids).                 |
+| `availableDataOriginalSquareSize` | `uint64`                  | The number of rows/columns of the original data [shares](data_structures.md#share) in [square layout](#arranging-available-data-into-shares) for this block. Must be a power of 2. |
```
musalbas, line 117: Why do we still need this field, isn't there a one-to-one mapping between availableDataOriginalSharesUsed and availableDataOriginalSquareSize?

specs/data_structures.md
musalbas: What is a compact root? This is the only time a reference to compact root is mentioned. How does an extra hash operation result in a compact root?

specs/data_structures.md
adlerjohn: `availableDataOriginalSharesUsed` is the number of shares used that aren't tail padding (up to `availableDataOriginalSquareSize ** 2`), and `availableDataOriginalSquareSize` is the size of the original square.

specs/data_structures.md
adlerjohn: The compact root is, as described, the hash of the NMT root (which is a tuple (min namespace ID, max namespace ID, hash). I should probably use the expression "compact root" explicitly, but they're used for the row and column roots in the available data header.

adlerjohn: Fixed in ccffc28.

specs/data_structures.md
musalbas: My point is that for any given value of `availableDataOriginalSharesUsed`, it seems to me that given value should always have the same `availableDataOriginalSquareSize`, so `availableDataOriginalSquareSize` can just be recomputed from `availableDataOriginalSharesUsed`.

adlerjohn: Assuming we have a rule that the square size must always be as small as possible, I think that's right. But we don't yet have such a rule. Do we want one?

musalbas: If we remove availableDataOriginalSquareSize, that will be implicit, yet.

adlerjohn: Removed `availableDataOriginalSquareSize` in 3a5c731. cc @liamsi

specs/data_structures.md
```diff
@@ -105,18 +105,19 @@ Blocks are the top-level data structure of the LazyLedger blockchain.
...
+| `availableDataOriginalSharesUsed` | `uint64`                  | The number of shares used in the [original data square](#arranging-available-data-into-shares) that are not [tail padding](./consensus.md#reserved-namespace-ids). |
+| `availableDataRoot`               | [HashDigest](#hashdigest) | Root of [commitments to erasure-coded data](#availabledataheader).                                                                                                 |
+| `proposerAddress`                 | [Address](#address)       | Address of this block's proposer.                                                                                                                                  |
+
+The size of the [original data square](#arranging-available-data-into-shares), `availableDataOriginalSquareSize`, isn't explicitly declared in the block header. Instead, it is implicitly computed as the smallest power of 2 whose square is at least `availableDataOriginalSharesUsed` (in other words, the smallest power of 4 that is at least `availableDataOriginalSharesUsed`).
```
liamsi, line 120: Makes sense.

liamsi (approved):
LGTM

PR #99, opened 2020-12-02
Add fees with fee burning

adlerjohn:
See the rationale document for rationale: https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-fees/rationale/fees.md

Add consensus rules and data structures for paying fees.


rationale/fees.md
```diff
@@ -0,0 +1,36 @@
...
+
+From this, we see that there is nothing inherent in half-fullness, and indeed this parameter should be chosen more carefully through empirical results.
+
+## Fee Burning
+
+Since blocks grow in size by a power of 4 due to the arrangement of data in a square, a more reasonable initial choice of parameters would be 1:4 target:maximum block size limits. We can then define the base fee as
+
+$$
+b_{cur} = b_{prev} \cdot \left( 1 + R \cdot \frac{s_{prev} - S_{target}}{S_{target}} \right)
+$$
```
liamsi, line 32: rendered (for reviewers like me whose brain doesn't immediately translate this into a readable formula):
image

specs/architecture.md
```diff
@@ -9,16 +9,16 @@ Architecture
...
 | App (application) | Alternate name for "[virtual sidechain](https://arxiv.org/abs/1905.09274)." LazyLedger apps are sidechains that post all their data onto the LazyLedger chain to share security. |
-| Transaction       | Requests that modify the consensus-critical state (validator balances and statuses).                                                                                             |
-| Message           | Requests that are executed by non-consensus-critical apps.                                                                                                                       |
+| Transaction       | Request that modifies the consensus-critical state (validator balances and statuses).                                                                                            |
+| Message           | Request that is executed by a non-consensus-critical app.                                                                                                                        |
```
liamsi, line 13: Didn't we want to move away from the word Request here?

adlerjohn: I don't think so? It's a term taken from the Tendermint paper. If there's an alternative for "atomic blob that is processed by the state machine," then I can change it.

specs/consensus.md
```diff
@@ -176,30 +179,51 @@ For `wrappedTransaction`'s [transaction](./data_structures.md#transaction) `tran
 
 1. `transaction.signature` must be a [valid signature](./data_structures.md#public-key-cryptography) over `transaction.signedTransactionData`.
 
-Finally, each `wrappedTransaction` is processed depending on [its transaction type](./data_structures.md#signedtransactiondata). These are specified in the next subsections, where `tx` is short for `transaction.signedTransactionData`, and `sender` is the recovered signing [address](./data_structures.md#address). After applying a transaction, the new state state root is computed.
+Finally, each `wrappedTransaction` is processed depending on [its transaction type](./data_structures.md#signedtransactiondata). These are specified in the next subsections, where `tx` is short for `transaction.signedTransactionData`, and `sender` is the recovered signing [address](./data_structures.md#address). We will define a helper functions:
```
liamsi, line 182:
```suggestion
Finally, each `wrappedTransaction` is processed depending on [its transaction type](./data_structures.md#signedtransactiondata). These are specified in the next subsections, where `tx` is short for `transaction.signedTransactionData`, and `sender` is the recovered signing [address](./data_structures.md#address). We will define a few helper functions:
```

specs/consensus.md
```diff
@@ -176,30 +179,51 @@ For `wrappedTransaction`'s [transaction](./data_structures.md#transaction) `tran
...
+```
+baseCost(y) = block.header.feeHeader.baseRate * y
+tipCost(y) = block.header.feeHeader.tipRate * y
+totalCost(x, y) = x + baseCost(y) + tipCost(y)
```
liamsi, line 186: Would it make sense to state the domain of these functions here?

rationale/fees.md
```diff
@@ -0,0 +1,36 @@
...
+- [Background: EIP-1559](#background-eip-1559)
+- [Fee Burning](#fee-burning)
+
+## Preamble
+
+Two desiderata are elastic block size limits—which can smooth out transaction fees and prevent fee spikes—and variable burning of fees—which can respond to market forces and provides intrinsic demand for the burned coin that cannot be replicated by issuance of new supply. Interestingly, these two properties cannot be accomplished independently. Without a negative feedback mechanism that cannot be bypassed, block producers will always produce maximum-size blocks. This negative feedback mechanism is in-protocol variable fee burning. In a similar vein, variable (rather than fixed) fee burning requires a measure of block usage, which happens to exactly be an elastic block size limit.
+
+This document describes the rationale and structure of an elastic block size limit through variable in-protocol fee burning.
+
+## Background: EIP-1559
```
liamsi, line 14: Shouldn't this section mention if we do anything different from EIP-1559? I was not familiar with that eip and it's not obvious if there is anything we do really differently here (besides renaming).

specs/data_structures.md
```diff
@@ -170,6 +172,24 @@ HashDigest is a [type alias](#type-aliases).
 
 Output of the [hashing](#hashing) function. Exactly 256 bits (32 bytes) long.
 
+### FeeHeader
+
+| name       | type     | description                                      |
+| ---------- | -------- | ------------------------------------------------ |
+| `baseRate` | `uint64` | The base fee rate for this block.                |
```
liamsi, line 179: Why not call it baseFee if it is the base fee?

rationale/fees.md
adlerjohn: The only thing that's different here from EIP-1559 is

> Since blocks grow in size by a power of 4 due to the arrangement of data in a square, a more reasonable initial choice of parameters would be 1:4 target:maximum block size limits.

which is mentioned further down. Is that sufficient?

specs/architecture.md
liamsi: > "atomic blob that is processed by the state machine," 

Transaction?

specs/consensus.md
adlerjohn: Sure.

specs/architecture.md
adlerjohn: Okay, well, it's outside the scope of this PR. I'll file an issue to follow up on potentially better nomenclature for this. #101

specs/data_structures.md
adlerjohn: Because it's not a fee, it's a rate (per byte). Similarly to how "gas price" isn't a fee, but a rate.

specs/data_structures.md
```diff
@@ -112,6 +113,7 @@ Block header, which is fully downloaded by both full clients and light clients.
 | `lastBlockID`                     | [BlockID](#blockid)       | Previous block's ID.                                                                                                                                               |
 | `lastCommitHash`                  | [HashDigest](#hashdigest) | Previous block's Tendermint commit hash.                                                                                                                           |
 | `consensusRoot`                   | [HashDigest](#hashdigest) | Merkle root of [consensus parameters](#consensus-parameters) for this block.                                                                                       |
+| `feeHeader`                       | [FeeHeader](#feeheader)   | Header data pertaining to fees.                                                                                                                                    |
```
liamsi, line 116: The rationale to not include this in the header is because it isn't data light clients would need to care about?

liamsi (commented):
I didn't fully familiarize myself with eip-1559 but the changes make sense. Also, I didn't yet validate each check and each state transition. The ones I've checked make sense so far and besides some minor comments the PR looks good to me.

specs/data_structures.md
adlerjohn: Oh no. It's abstracted like this in case we want to e.g. not do fee burning. Then the fields pertaining to fees can be swapped out without affecting everything else.

rationale/fees.md
liamsi: Not sure. Yes, from a spec-only perspective, it is sufficient. From an accessibility pov, it would help someone familiar with eip-1559 to immediately grasp how we do fee burning to have a sentence (somewhere) stating that besides that minor difference they are essentially the same.

specs/consensus.md
adlerjohn: Fixed in 4a951c6.

rationale/fees.md
adlerjohn: Clarified further in 2e01d72.

liamsi (approved):
LGTM! I think we'll give this a second and much more detailed pass, when triying to implement this (e.g. as a cosmos module - this discussion is kinda related btw: https://github.com/cosmos/cosmos-sdk/issues/7347)

Issue #83, opened 2020-10-13
Investigate different hash functions


adlerjohn:
This issue is to track investigation into which hash function to use throughout the primary LazyLedger data structures (i.e. Merkle trees and simple hashes).

Two critical desiderata:
1. **Standard**: Using a nonstandard cryptographic primitive can lead to [long-term widespread confusion](https://github.com/ethereum/EIPs/issues/59) and should be avoided at all costs.
1. **Ethereum compatibility**: Ethereum has native support for the following hash algorithms: `keccak256`, `sha256`, `blake2b`, `ripemd160`.

Note that performance is a secondary concern at best.

Of Ethereum's natively supported hash algorithms, only `sha256` is standard (i.e. approved by a standards body). However, it is vulnerable to a [length extension attack](https://en.wikipedia.org/wiki/Length_extension_attack). The severity of this should be considered low both because it only applies to the use of this hash function in a MAC, and because the extensive use of `sha256` in the cryptocurrency ecosystem serves as a "crowd immunity" layer.

[SHA-3](https://en.wikipedia.org/wiki/SHA-3) would be a preferable alternative to avoid this issue altogether, but is not natively supported by Ethereum at this time.

PR #103, opened 2020-12-14
Change hash function to SHA-2-256

adlerjohn:
As per #83.


liamsi (approved):
LGTM

adlerjohn:
Don't think there are any. Made sure to abstract the choice of hashing function to just this section 😂

PR #105, opened 2021-01-03
Add end block state transition

adlerjohn:
Represent the validator queue as a singly linked list. This allows us to collect the top `MAX_VALIDATORS` validators in linear time to the `MAX_VALIDATORS`, independent of the number of queued validators. Each modification to the validator queue can be done in logarithmic time w.r.t. the number of queued validators if a local tree data structure is used to store the queue (which clients are expected to actually do). In other words, the linked list representation is primarily for light clients/fraud proofs.

Add an end block state implicit transition to make use of the queue to collect the top `MAX_VALIDATORS` trivially by walking through the linked list.

Also refactored the layout of delegations in the state tree just to make things a bit simpler. Instead of storing the delegation data with each account, store it in a dedicated delegations subtree.


adlerjohn:
Note: markdown-link-check is failing with this link https://www.secg.org/sec1-v2.pdf which is a bit unstable for me (sometimes fails to load) but isn't dead.

specs/consensus.md
```diff
@@ -194,6 +195,12 @@ totalCost(x, y) = x + baseCost(y) + tipCost(y)
 ```
 , where `x` above represents the amount of coins sent by the transaction authorizer and `y` above represents the a measure of the block space used by the transaction (i.e. size in bytes).
 
+Two additional helper functions are defined:
+1. `findFromQueue(power)`, which returns the address of the last validator in the [validator queue](./data_structures.md#validator) with voting power greater than or equal to `power`, or `0` if the queue is empty or no validators in the queue have at least `power` voting power.
+1. `parentFromQueue(address)`, which returns the address of the parent in the validator queue of the validator with address `address`, or `0` if `address` is not in the queue or is the head of the queue.
+
+Note that light clients cannot perform a linear search through a linked list, and are instead provided logarithmic proofs (e.g. in the case of `parentFromQueue`, a proof to the parent is provided, which should have `address` as its next validator).
+
```
liamsi, line 203: > a proof to the parent is provided, which should have `address` as its next validator

These are just Merkle proofs? I did not go through the state transition fraud proofs but a light client will need two proofs then (for the validator with `address` and a proof to the parent in the queue?

adlerjohn: Yes, they're just Merkle proofs, and yes, (up to) two would be needed any time a light client is executing `parentFromQueue` (e.g. in a fraud proof). Note that this isn't a different fraud proof type; it's just how the "pre-state" is given is a state transition fraud proof.

specs/data_structures.md
```diff
@@ -773,59 +774,54 @@ The state of the LazyLedger chain is intentionally restricted to containing only
...
-In the accounts subtree, accounts (i.e. leaves) are keyed by the [hash](#hashdigest) of their [address](#address). The first byte is then replaced with `ACCOUNTS_SUBTREE_ID`.
-
-### Delegation
-
 ```C++
-enum DelegationStatus : uint8_t {
-    Bonded = 1,
-    Unbonding = 2,
+enum AccountStatus : uint8_t {
+    None = 1,
```
liamsi, line 787: While the other states are more or less self-explanatory, it is not clear what this stands for. Maybe add a comment here?

specs/consensus.md
```diff
@@ -456,17 +509,43 @@ validator.pendingRewards = 0
...
+    if parent != 0
+        if state.accounts[parent].status == AccountStatus.ValidatorBonded
+            validator.next = state.activeValidatorSet[parent].next
+            state.activeValidatorSet[parent].next = delegation.validator
+        else
+            validator.next = state.inactiveValidatorSet[parent].next
+            state.inactiveValidatorSet[parent].next = delegation.validator
+    else
+        validator.next = state.validatorQueueHead
+        state.validatorQueueHead = delegation.validator
```
liamsi, line 537: Isn't this the exact same pseudo-code as in lines 438 following?

specs/data_structures.md
adlerjohn: Fixed in 2da3d9b.

specs/consensus.md
adlerjohn: Fixed in 33d1f04.

specs/data_structures.md
```diff
@@ -800,6 +800,18 @@ enum AccountStatus : uint8_t {
 | `nonce`   | [Nonce](#type-aliases)  | Account nonce. Every outgoing transaction from this account increments the nonce. |
 | `status`  | `AccountStatus`         | Validator or delegation status of this account.                                   |
 
+The `status` of an account indicates weather it is a validator (`AccountStatus.Validator*`), delegating to a validator (`AccountStatus.Delegation*`), or neither (`AccountStatus.None`). Being a validator and delegating are mutually exclusive, and only a single validator can be delegated to.
+
```
liamsi, line 804: Are the implications of this mentioned somewhere too? E.g., that you need to create another account per delegation (and that this doesn't cause the UX to get worse as the wallet/UI can simply use a different derivation path per delegation etc)?

specs/consensus.md
```diff
@@ -195,9 +195,41 @@ totalCost(x, y) = x + baseCost(y) + tipCost(y)
 ```
 , where `x` above represents the amount of coins sent by the transaction authorizer and `y` above represents the a measure of the block space used by the transaction (i.e. size in bytes).
 
-Two additional helper functions are defined:
+Four additional helper functions are defined to manage the [validator queue](./data_structures.md#validator):
 1. `findFromQueue(power)`, which returns the address of the last validator in the [validator queue](./data_structures.md#validator) with voting power greater than or equal to `power`, or `0` if the queue is empty or no validators in the queue have at least `power` voting power.
 1. `parentFromQueue(address)`, which returns the address of the parent in the validator queue of the validator with address `address`, or `0` if `address` is not in the queue or is the head of the queue.
+3. `validatorQueueInsert`, defined as
+```py
```
liamsi, line 202: minor nit: the other pseudo code comes without the `py`. This prob. does have only very minor implications on how the syntax is highlighted anyways but it would be good if this was consistent throughout the spec.

specs/data_structures.md
adlerjohn: No. Not sure if that belongs directly in the technical specs, but if it's something that comes up we can probably write a rationale document around our staking design later.

specs/consensus.md
adlerjohn: Fixed in dba9e99, noting that the pseudocode isn't in python but just uses that for decent syntax highlighting.

specs/consensus.md
```diff
@@ -101,6 +102,8 @@ The validity of a newly-seen block, `block`, is determined by two components, de
 1. [Block structure](#block-structure): whether the block header is valid, and data in a block is arranged into a valid and matching data root (i.e. syntax).
 1. [State transition](#state-transitions): whether the application of transactions in the block produces a matching and valid state root (i.e. semantics).
 
+Pseudocode in this section is not in any specific language and should be interpreted as being in a neutral and sane language.
```
musalbas, line 105: Is it at least consistent across the spec?

specs/consensus.md
```diff
@@ -187,18 +190,56 @@ For `wrappedTransaction`'s [transaction](./data_structures.md#transaction) `tran
...
+            validator.next = state.activeValidatorSet[parent].next
+            state.activeValidatorSet[parent].next = sender
+        else
+            validator.next = state.inactiveValidatorSet[parent].next
+            state.inactiveValidatorSet[parent].next = sender
+    else
+        validator.next = state.validatorQueueHead
+        state.validatorQueueHead = sender
+```
+4. `validatorQueueRemove`, defined as
```
musalbas, line 219: How do these functions translate to modifying the key-value store in the state? They seem too abstract for that?

specs/consensus.md
```diff
@@ -187,18 +190,56 @@ For `wrappedTransaction`'s [transaction](./data_structures.md#transaction) `tran
...
+    if parent != 0
+        if state.accounts[parent].status == AccountStatus.ValidatorBonded
+            state.activeValidatorSet[parent].next = validator.next
+            validator.next = 0
+        else
+            state.inactiveValidatorSet[parent].next = validator.next
+            validator.next = 0
+    else
+        state.validatorQueueHead = validator.next
+        validator.next = 0
```
musalbas, line 233: If this is to translate to the SMT/key-value store, I would've expected the relevant state key to be deleted?

specs/consensus.md
adlerjohn: Yes, it should all be consistent.

musalbas: It would probably be better to use Python syntax across the spec, since it's almost identical but it's a real language so there's a point of reference

adlerjohn: It might indeed be a good idea to make the code parts actually executable, and probably in python. This would also tie in nicely to building a set of test vectors for the spec. But we can leave that for a future issue.

specs/consensus.md
adlerjohn: They modify the queue head and the local validator object, which is the calling function is then saved to state.

specs/consensus.md
adlerjohn: The calling function actually handles deleting. This function just removes from the linked list without deleting, which is important because updates are abstracted as "remove followed by insert."

musalbas (approved):

Issue #101, opened 2020-12-08
Nomenclature of "request" is confusing


adlerjohn:

liamsi:
I think Request is very common in the PBFT/Consensus focused literature. See, e.g., http://pmg.csail.mit.edu/papers/osdi99.pdf 

I think in blockchain-land, the word Transaction is predominantly used for the same thing.

PR #111, opened 2021-01-19
Define node types and configuration parameters

### LINKED ISSUE #22: Define different node types

adlerjohn:
Fixes #22.

Define different node types as a function of some configuration parameters (different levels of resource requirements for block headers, block bodies, and transactions).

### DIFF
--- a/specs/node_types.md
+++ b/specs/node_types.md
@@ -99,7 +99,7 @@ For convenience, we will define several common parameter configurations:
     - Block headers: [Extended Block Headers](#extended-block-headers)
     - Block bodies: [Sampled Bodies](#sampled-bodies)
     - Transactions: [Full Transactions](#full-transactions)
-1. Storage nodes provide the same security guarantees as full nodes. Block bodies are stored and served to the network.
+1. Storage nodes provide the same security guarantees as full nodes. Block bodies (in [erasure-coded form](./data_structures.md#2d-reed-solomon-encoding-scheme)) are stored and served to the network.
     - Block headers: [Extended Block Headers](#extended-block-headers)
     - Block bodies: [Full Bodies](#full-bodies)
     - Transactions: [Full Transactions](#full-transactions)
--- a/specs/node_types.md
+++ b/specs/node_types.md
@@ -1,0 +1,4 @@
+Node Types
+===
+
+
--- a/specs/node_types.md
+++ b/specs/node_types.md
@@ -19,7 +19,7 @@ Node Types
 
 Nodes that run the [LazyLedger protocol](./consensus.md) have a number of parameters that can be tweaked with regards to which parts of the data is downloaded, validated, and/or stored. All nodes process the `header` and `lastCommit` fields of each [block](./data_structures.md#block), but can handle the `availableDataHeader` and `availableData` fields differently.
 
-We define security assumptions as assumptions under which a given node is guaranteed accountable consensus safety (i.e. that finalized blocks will remain in the prefix of all future sequences of blocks accepted by the node, unless a supermajority of validator voting power performs an attributable—and thus [penalizable](./consensus.md#blockavailabledataevidencedata)—malicious action) and state safety (i.e. that an invalid state transition will not be included in the chain accepted by the node).
+We define security assumptions as assumptions under which a given node is guaranteed accountable consensus safety (i.e. that finalized blocks will remain in the prefix of all future sequences of blocks accepted by the node, unless a supermajority (> 2/3) of validator voting power performs an attributable—and thus [penalizable](./consensus.md#blockavailabledataevidencedata)—malicious action) and state safety (i.e. that an invalid state transition will not be included in the chain accepted by the node).
 
 ### Block Headers
 
--- a/specs/node_types.md
+++ b/specs/node_types.md
@@ -78,4 +78,28 @@ At most secure under a weak subjectivity assumption.
 
 ## Node Types
 
-[full node](https://en.bitcoin.it/wiki/Full_node)
+For convenience, we will define several common parameter configurations:
+1. [Full nodes](https://en.bitcoin.it/wiki/Full_node) provide the strongest security guarantees. Block bodies are not stored.
+    - Block headers: [Extended Block Headers](#extended-block-headers)
+    - Block bodies: [Full Bodies](#full-bodies)
+    - Transactions: [Full Transactions](#full-transactions)
+1. Partial nodes are capable of producing fraud proofs of invalid transactions and contribute to validating the erasure coding of random blocks.
+    - Block headers: [Extended Block Headers](#extended-block-headers)
+    - Block bodies: [Partial Bodies](#partial-bodies)
+    - Transactions: [Full Transactions](#full-transactions)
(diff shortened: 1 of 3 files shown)


specs/node_types.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Block Bodies
+
+Block bodies (the `availableData` [block](./data_structures.md#block) field) can be downloaded and optionally stored and/or served. Storing and serving block body data has no effect on node security assumptions.
+
+#### No Bodies
+
+Nodes that only process [compact block headers](#compact-block-headers) have no need for block bodies and simply do not process block bodies.
+
+Secure under an honest supermajority majority of validator voting power and a weak subjectivity assumption.
```
liamsi, line 45:
```suggestion
Secure under an honest supermajority majority (e.g. >= 2/3) of validator voting power and a weak subjectivity assumption.
```

specs/node_types.md
```diff
@@ -0,0 +1,105 @@
+Node Types
+===
+
+- [Node Parameters](#node-parameters)
```
liamsi, line 4: IMO, it would be good to summarize the different node types and their security in some kind of matrix or table and add it to a section called Overview.

specs/node_types.md
```diff
@@ -0,0 +1,105 @@
...
+
+These nodes fully download and validate [the erasure coding](./data_structures.md#2d-reed-solomon-encoding-scheme) of all block bodies.
+
+If [transactions are not processed](#no-transactions), secure under an honest minority of nodes and a weak subjectivity assumption. If [transactions are processed](#full-transactions), secure under a weak subjectivity assumption.
+
+### Transactions
+
+#### No Transactions
+
+These nodes process do not process requests [with a reserved namespace ID](./data_structures.md#arranging-available-data-into-shares) and thus to not know the chain state without relying on a third party.
```
liamsi, line 69: Didn't we want to use transactions in favor of "requests"? #101

specs/node_types.md
```diff
@@ -0,0 +1,105 @@
...
+
+At most secure under an honest minority of nodes and a weak subjectivity assumption.
+
+#### Full Transactions
+
+Nodes that wish to produce new blocks must know the [chain state](./data_structures.md#state). Processing all block bodies is actually not needed to know the LazyLedger state, as [transactions that pay for message inclusion commit to messages](./../rationale/message_block_layout.md). These nodes process all requests [with a reserved namespace ID](./data_structures.md#arranging-available-data-into-shares) from block bodies and perform Data Availability Sampling for the remaining (message) data.
```
liamsi, line 75: #101

specs/node_types.md
liamsi: NVM, I see there is a `## Node Types` at the bottom which is great!

specs/node_types.md
```diff
@@ -0,0 +1,105 @@
...
+
+At most secure under a weak subjectivity assumption.
+
+## Node Types
+
+For convenience, we will define several common parameter configurations:
+1. [Full nodes](https://en.bitcoin.it/wiki/Full_node) provide the strongest security guarantees. Block bodies are not stored.
```
liamsi, line 82: > Block bodies are not stored.

But they can be?

specs/node_types.md
```diff
@@ -0,0 +1,105 @@
...
+    - Block headers: [Extended Block Headers](#extended-block-headers)
+    - Block bodies: [Full Bodies](#full-bodies)
+    - Transactions: [Full Transactions](#full-transactions)
+1. Partial nodes are capable of producing fraud proofs of invalid transactions and contribute to validating the erasure coding of random blocks.
+    - Block headers: [Extended Block Headers](#extended-block-headers)
+    - Block bodies: [Partial Bodies](#partial-bodies)
+    - Transactions: [Full Transactions](#full-transactions)
+1. Light nodes DAS and are secure under an honest minority.
```
liamsi, line 90: It's the first time the abbreviation is used (DAS).

specs/node_types.md
```diff
@@ -0,0 +1,105 @@
...
+    - Transactions: [No Transactions](#no-transactions)
+1. Superlight nodes do not perform DAS and are secure under an honest majority.
+    - Block headers: [Compact Block Headers](#compact-block-headers)
+    - Block bodies: [No Bodies](#no-bodies)
+    - Transactions: [No Transactions](#no-transactions)
+1. Light validator nodes can produce new blocks with strong security guarantees and light resource requirements.
+    - Block headers: [Extended Block Headers](#extended-block-headers)
+    - Block bodies: [Sampled Bodies](#sampled-bodies)
+    - Transactions: [Full Transactions](#full-transactions)
+1. Storage nodes provide the same security guarantees as full nodes. Block bodies are stored and served to the network.
```
liamsi, line 102: Would you make a difference between (1) a node that stores and serves the raw availableData and (2) a node that stores and serves the erasure-coded version of that data (e.g. an ipfs node that serves the rows and columns)? Or would you rather say, it's the same thing and just a different encoding? 

The interfaces to interact with these nodes is quite different though:
 - (1) give me block data of block with hash x / height x
 - (2) give me leaf-data of the erasure-block with hash x with index y (or path, cid/0/0/1/0/0), or, give me all leaves of namespace nid (could also be a path root/nid)

specs/node_types.md
adlerjohn: Any user is free to configure these parameters differently, but the expression "full nodes" should be reserved for the node type that only keep track of the latest state and not the entire history. It's been a point of confusion with other blockchains that we should draw the line on.

specs/node_types.md
adlerjohn: I'd rather subsume all changes to the word "request" into a single PR and for now follow along with the same expression consistently.

specs/node_types.md
adlerjohn: Clarified in a5efbe1.

specs/node_types.md
adlerjohn: Nonetheless, clarified wording in e5d2320.

specs/node_types.md
adlerjohn: I don't really see a need to distinguish the two cases. It's two different encodings of the same data.

But technically, when placed in a 2D square, the available data actually gains information in the form of specific placement of shares that isn't captured with the raw available data (since the block proposer can include padding shares). As such, storage nodes should probably be storing and serving the erasure-coded version of the block bodies.

specs/node_types.md
liamsi: OK

specs/node_types.md
liamsi: Should we then change this to
```suggestion
1. Storage nodes provide the same security guarantees as full nodes. Block bodies (in [erasure-coded form](./data_structures.md#2d-reed-solomon-encoding-scheme)) are stored and served to the network.
```
?

liamsi (approved):
LGTM 🚀

Issue #108, opened 2021-01-15
Using Python syntax as pseudocode


musalbas:
I think Python syntax should be used a pseudocode, as it's almost identical to the current (fake) syntax, and since it's a real language it provides a point of reference.

PR #112, opened 2021-01-19
Use raw data for Merkle tree inner nodes instead of serialization

### LINKED ISSUE #109: NMT clarify leaves hashing
It’s not clear if the namespace is also fed into the hash `h` as this might or might not be the case depending on what the `serialize` function looks like. I though I remembered that `serialize` was narrowed down by specifiying if serialization contains the NID or not but I can't find it currently. In the current form the spec would allow both cases as far as I understand. From a security pov both should be fine as the NID is returned / prefixed in front of the hash anyways (and hence the tree commits to the namespace of each leaf too).

> For leaf node `node` of data `d`:
> ```C++
> node.n_min = d.namespaceID
> node.n_max = d.namespaceID
> node.v = h(0x00, serialize(d))
> ```
> 

https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#namespace-merkle-tree

In the implementation it is not and a leaf hash is computed as: 
`ns(rawData) || ns(rawData) || hash(leafPrefix || rawData)`  where `rawData` does not contain the namespace.
or more precisely:
https://github.com/lazyledger/nmt/blob/4b4c58a9e7c071db650b27a69dccde4821411d07/internal/hasher.go#L63-L67

----------------
IMHO, this is a perfect candidate to start putting https://github.com/lazyledger/lazyledger-specs/issues/108 into practice.

adlerjohn:
Fixes #109.

Instead of serializing child nodes (which would require the use of a single serialization scheme but introduce a dependency on a serialization library to the trees), just use the raw bytes (which is technically a different serialization scheme but requires no library).

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -284,7 +284,7 @@ node.v = h(0x00, serialize(d))
 
 For internal node `node` with children `l` and `r`:
 ```C++
-node.v = h(0x01, serialize(l), serialize(r))
+node.v = h(0x01, l.v, r.v)
 ```
 
 Note that rather than duplicating the last node if there are an odd number of nodes (the [Bitcoin design](https://github.com/bitcoin/bitcoin/blob/5961b23898ee7c0af2626c46d5d70e80136578d3/src/consensus/merkle.cpp#L9-L43)), trees are allowed to be imbalanced. In other words, the height of each leaf may be different. For an example, see Section 2.1.3 of [Certificate Transparency (RFC-6962)](https://tools.ietf.org/html/rfc6962#section-2.1.3).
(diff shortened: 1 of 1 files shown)


liamsi:
Can we use python as pseudo code (#108) while touching these lines? It will further disambiguate things.

adlerjohn:
> Can we use python as pseudo code (#108) while touching these lines?

I'd like to keep work on #108 self-contained to a PR dedicated for that and do a whole sweep of all code snippets at once.

specs/data_structures.md
```diff
@@ -320,11 +320,11 @@ node.n_max = 0x0000000000000000000000000000000000000000000000000000000000000000
 node.v = 0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
 ```
 
-For leaf node `node` of data `d`:
+For leaf node `node` of [share](#share) data `d`:
 ```C++
 node.n_min = d.namespaceID
 node.n_max = d.namespaceID
-node.v = h(0x00, serialize(d))
+node.v = h(0x00, d.rawData)
```
liamsi, line 327: OK, that is the same as the current implementation 👍🏼

liamsi (approved):
👍🏼

Issue #104, opened 2020-12-16
Consider adding rewards to voting power


adlerjohn:
Currently, rewards are held as pending rewards, but aren't added as voting power. This necessitates validators unbond completely to fold their rewards into their voting power, which is poor UX. We should consider immediately doing that automatically, and investigate if this leads to any potential issues.

PR #114, opened 2021-01-26
Add delegation rewards and commission

### LINKED ISSUE #95: Handle delegation commission

adlerjohn:
Fixes #95, #104.

Major changes:
* Remove `stakedBalance` for validators. Validators now have no balance and need to delegate to themselves. This can be abstracted away at the UI layer.
* Delegation commission is computed and rewards (both block subsidy and fees) are dispersed to both validators and delegations. Commission rewards are stored in a new field in the validator object.

Currently validator commission rewards simply sit there until the validator unbonds. We should support delegating this reward #117.

### DIFF
--- a/specs/consensus.md
+++ b/specs/consensus.md
@@ -393,9 +393,6 @@ state.accounts[sender].status = AccountStatus.ValidatorUnbonded
 
 state.accounts[sender].balance += validator.stakedBalance
 
-validator.votingPower -= validator.stakedBalance
-validator.stakedBalance = 0
-
 state.inactiveValidatorSet[sender] = validator
 
 if validator.delegatedCount == 0
--- a/specs/consensus.md
+++ b/specs/consensus.md
@@ -544,9 +544,10 @@ state.accounts[sender].nonce += 1
 state.accounts[sender].balance -= totalCost(0, bytesPaid)
 state.accounts[sender].status = None
 
+# Return the delegated stake
 state.accounts[sender].balance += delegation.stakedBalance
-postCommissionRewards = validator.commissionRate * delegation.stakedBalance * (delegation.endEntry - delegation.beginEntry)
-state.accounts[sender].balance += postCommissionRewards
+# Also disperse rewards (commission has already been levied)
+state.accounts[sender].balance += delegation.stakedBalance * (delegation.endEntry - delegation.beginEntry)
 
 if state.accounts[delegation.validator].status == AccountStatus.ValidatorQueued ||
       state.accounts[delegation.validator].status == AccountStatus.ValidatorUnbonding
@@ -616,8 +617,9 @@ else if account.status == AccountStatus.ValidatorBonded
     proposer = state.activeValidatorSet[block.header.proposerAddress]
 
 blockReward = state.activeValidatorSet.proposerBlockReward
-proposer.commissionRewards += proposer.commission * blockReward
-proposer.pendingRewards += blockReward
+commissionReward = proposer.commission * blockReward
+proposer.commissionRewards += commissionReward
+proposer.pendingRewards += blockReward - commissionReward
 
 # Even though the voting power hasn't changed yet, we consider this a period change.
 # Note: this isn't perfect because the block reward is shared with delegations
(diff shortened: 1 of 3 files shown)


specs/consensus.md
```diff
@@ -595,12 +596,47 @@ At the beginning of the block, rewards are distributed to the block proposer.
...
+proposer.votingPower += blockReward
+state.activeValidatorSet.activeVotingPower += blockReward
+
+if account.status == AccountStatus.ValidatorUnbonding
+      account.status == AccountStatus.ValidatorUnbonded
+    state.inactiveValidatorSet[block.header.proposerAddress] = proposer
+else if account.status == AccountStatus.ValidatorBonded
+    state.activeValidatorSet[block.header.proposerAddress] = proposer
+```
+
```
liamsi, line 639: Is there an equivalent of allocating rewards according to voting power?
e.g. this: https://github.com/cosmos/cosmos-sdk/blob/e2f510afcc3478edab4a7d8d786dccbec333dddd/x/distribution/keeper/allocation.go#L80-L95

liamsi (approved):
LGTM! 
I did not go through the state transition pseudo-code in detail though, as this part of the spec isn't immediately implementation critical, as we we will use the staking / distribution [modules](https://github.com/cosmos/cosmos-sdk/tree/master/x#list-of-modules) without any modifications at first.

specs/consensus.md
```diff
@@ -307,14 +308,13 @@ Apply the following to the state:
...
-state.accounts[sender].balance -= totalCost(tx.amount, bytesPaid)
+state.accounts[sender].balance -= totalCost(0, bytesPaid)
 state.accounts[sender].status = AccountStatus.ValidatorQueued
 
 validator = new Validator
-validator.stakedBalance = tx.amount
 validator.commissionRate = tx.commissionRate
 validator.delegatedCount = 0
-validator.votingPower = tx.amount
+validator.votingPower = 0
```
marbar3778, line 317: What is the benefit here?

specs/consensus.md
```diff
@@ -585,7 +586,7 @@ Apply the following to the state:
 state.accounts[sender].nonce += 1
 state.accounts[sender].balance -= totalCost(tx.amount, bytesPaid)
 
-state.activeValidatorSet[block.header.proposerAddress].pendingRewards += tipCost(bytesPaid)
+state.activeValidatorSet.proposerBlockReward += tipCost(bytesPaid)
 ```
 
 #### Begin Block
```
marbar3778, line 592: This spec assumes these computations will be done every block. Why not epoch it?

liamsi: I guess this is because this is how it works in the SDK currently. While we want epochs on the long run I'm against putting more effort into this now. It is "just" an optimization (yeah, a really cool one but I think there are higher prio ones).

marbar3778: https://github.com/cosmos/cosmos-sdk/discussions/8328 😉 Should land in the sdk before launch. May be worth leaving a note that epochs are preferred.

liamsi: Nice! Thanks for bringing this to our attention.

specs/consensus.md
adlerjohn: No, both the block subsidy and fees are given entirely to the proposer. Over the long run this averages out to the same thing as if you spread it out each block so I don't think it makes much difference.

specs/consensus.md
adlerjohn: This is part of the refactor of removing staked balance for validators:

> Remove `stakedBalance` for validators. Validators now have no balance and need to delegate to themselves. This can be abstracted away at the UI layer.

(Note that this does introduce a divide-by-zero case when computing entries, but I'll take care of that in a future PR.)

specs/consensus.md
liamsi: That makes sense intuitively and  sounds reasonable but might require further analysis (?). For a first implementation, we will use the distribution module's logic and later change that logic accordingly.

specs/consensus.md
adlerjohn: The only real benefit to epochs for our usecase is reducing light client overhead since fewer block headers need to be downloaded. I'll probably be important long-term, but it's just an optimization and we can worry about those later.

PR #118, opened 2021-02-02
Support delegation of commission and rewards without withdrawals

### LINKED ISSUE #117: Support redelegating validator commission
Validator commission rewards are currently only withdrawable by unbonding the validator. We should support validators assigning their commission to a delegation without unbonding.

This can be done by either:
1. A new transaction type. Note that delegating reward this way can be abstracted away at the UI layer.
1. Automatically.

adlerjohn:
Fixes #117.

* Tangential: Fix block reward to be distributed based on voting power at beginning of block, not end.
* Add two new tx types to "redelegate" pending rewards (for delegations) and pending commission rewards (for validators). Relegating requires moving all pending coins and computes new F1 entries since voting power changes.

### DIFF
--- a/specs/consensus.md
+++ b/specs/consensus.md
@@ -611,6 +611,8 @@ state.activeValidatorSet[block.header.proposerAddress] = proposer
 
 #### End Block
 
+Apply the following to the state:
+
 ```py
 account = state.accounts[block.header.proposerAddress]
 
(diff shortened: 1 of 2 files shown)


liamsi (approved):
Again, I have not validated the state transition pseudo-code. IMO, all this should be revisited when / before actually implementing this.

PR #124, opened 2021-02-10
Define non-integer type

### LINKED ISSUE #120: Define decimal type
Currently we use a placeholder "Decimal" type https://github.com/lazyledger/lazyledger-specs/blob/1a71da17e8faf7ca3c85d4624c6cbe29205bf1df/specs/data_structures.md#decimal that is supposed to represent a decimal type. This type needs to be defined.

Also see:
- https://github.com/cosmos/cosmos-sdk/blob/bb85a690abc78abec1588709ee884c4d33d799e3/types/decimal.go#L17
- https://github.com/cosmos/cosmos-sdk/issues/7773

adlerjohn:
Fixes #120.

Defines the Decimal non-integer type as a rational number.

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -930,7 +930,12 @@ For explanation on entries, see the [reward distribution rationale document](../
 
 ### Decimal
 
-TODO define a format for numbers in the range `[0,1]`
+| name          | type   | description           |
+| ------------- | ------ | --------------------- |
+| `numerator`   | uint64 | Rational numerator.   |
+| `denominator` | uint64 | Rational denominator. |
+
+Represents a (potentially) non-integer number.
 
 ## Consensus Parameters
 
--- a/specs/consensus.md
+++ b/specs/consensus.md
@@ -722,7 +722,7 @@ proposer.latestEntry += proposer.pendingRewards // proposer.votingPower
 proposer.pendingRewards = 0
 
 blockReward = state.activeValidatorSet.proposerBlockReward
-commissionReward = proposer.commission * blockReward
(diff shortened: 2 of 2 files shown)


adlerjohn:
@liamsi is there anything to worry about w.r.t. overflowing on the multiplication? I assume that a simple "if it overflows, the tx is invalid" check is sufficient?

specs/data_structures.md
```diff
@@ -930,7 +930,12 @@ For explanation on entries, see the [reward distribution rationale document](../
 
 ### Decimal
 
-TODO define a format for numbers in the range `[0,1]`
+| name          | type   | description           |
+| ------------- | ------ | --------------------- |
+| `numerator`   | uint64 | Rational numerator.   |
+| `denominator` | uint64 | Rational denominator. |
+
+Represents a (potentially) non-integer number.
```
liamsi, line 938: This will be a relatively big change compared to the SDK. I guess switching from the current implementation to this rational type is really low priority, right?

adlerjohn: Right, it doesn't affect the devnet at all. Both Decimal as it currently exists in the SDK and Rational (i.e. our Decimal type) would produce similar numbers, so it wouldn't even change the state transitions in anything but in some cases that required lots of precision.

liamsi: Notes under  "optional to implement until we there is nothing left to crunch on" :-D  (tagging @evan-forbes s.t. he is also aware of this.

liamsi (approved):
👍🏼

PR #125, opened 2021-02-10
Handle divide-by-zero when computing new latest entries

adlerjohn:
When the voting power is 0, just return 0 as the entry. This may happen if all a validator's delegations undelegate, leaving them with 0 voting power.

### DIFF
--- a/specs/consensus.md
+++ b/specs/consensus.md
@@ -744,7 +744,7 @@ else if account.status == AccountStatus.ValidatorBonded
     state.activeValidatorSet[block.header.proposerAddress] = proposer
 ```
 
-At the end of a block, the top `MAX_VALIDATORS` validators by voting power are or become active (bonded). For newly-bonded validators, the entire validator object is moved to the active validators subtree and their status is changed to bonded. For previously-bonded validators that are no longer in the top `MAX_VALIDATORS` validators begin unbonding.
+At the end of a block, the top `MAX_VALIDATORS` validators by voting power with voting power _greater than_ zero are or become active (bonded). For newly-bonded validators, the entire validator object is moved to the active validators subtree and their status is changed to bonded. For previously-bonded validators that are no longer in the top `MAX_VALIDATORS` validators begin unbonding.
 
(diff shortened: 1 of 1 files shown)


adlerjohn:
Q @liamsi: would it be better/a good idea independently to add a check during end block to ensure that a validator with 0 voting power can't be in the validator set, even if there are fewer than the maximum validators with non-zero voting power?

liamsi:
That sounds reasonable. Wouldn't an actual validator with voting power 0 be something that can't happen and if it somehow happend that would be a serious bug in the state machine?

liamsi:
If Tendermint receives a validator with the power 0 from the app, it will remove the validator from the validator set:
https://github.com/lazyledger/lazyledger-core/blob/585566317e519bbb6d35d149b7e856c4c1e8657c/types/validator_set.go#L397
(calling [code](https://github.com/lazyledger/lazyledger-core/blob/585566317e519bbb6d35d149b7e856c4c1e8657c/state/execution.go#L452-L468)).

adlerjohn:
> Wouldn't an actual validator with voting power 0 be something that can't happen and if it somehow happend that would be a serious bug in the state machine?

The specs don't currently enforce this, but I think they should. It's a reasonable invariant to have that will probably save us some weird corner cases.

liamsi:
As long as it is still possible to pass 0-power validators through the ABCI boundary (to tell tendermint/lazyledger-core to remove them from the valset it's tracking), that should be ok.

liamsi:
As a note, these 0-power "validators" are created during endblock in the sdk, here: https://github.com/cosmos/cosmos-sdk/blob/cb725e1959fc03c58032301b8dc476b9de97ddc5/x/staking/keeper/val_state_change.go#L200

adlerjohn:
Hmm. Actually, this is a bit more tricky, because it's possible for a bonded validator to have zero voting power: if they had non-zero voting power and then all their delegations undelegated. So the invariant would only be on new validators joining the validator set, not in all cases.

liamsi:
> So the invariant would only be on new validators joining the validator set, not in all cases.

That makes sense (and is enforced in tendermint too - you can't create a new avlidator set that contains a val with zero votingpower).

adlerjohn:
Should be good now @liamsi

liamsi (approved):
👏🏼

PR #127, opened 2021-02-16
Define consensus messages

### LINKED ISSUE #126: Consensus Messages
We need to define the messages that are exchanged by nodes during consensus. 

IMO, most of the messages defined by tendermint can be used and do not require any changes. The main change would be regarding the [BlockID](https://docs.tendermint.com/master/spec/core/data_structures.html#blockid) (which is used in [Votes](https://docs.tendermint.com/master/spec/core/data_structures.html#vote), [Proposal](https://docs.tendermint.com/master/spec/core/data_structures.html#proposal) etc).

Note that we also have a [BlockId](https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#blockid) in the ll-spec which is the hash of the header basically (differently from tendermint and still different from what would be required for validators s.t. can sample for DA proofs).

My understanding is that the Proposal should contain the data root at least. IMO, it should also contain the [DA header ](https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#availabledataheader) and the orig data square size. The last is optionally but it would simplify the implementation (and I don't see why it should not simply be included).

Happy to help with this as we need to define these for the implementation anyway!

adlerjohn:
Fixes #126.

- Define _proposals_ and _votes_.
- Rename block ID to header hash, and hash header instead of Merkleizing it ([ref](https://github.com/tendermint/spec/issues/175#issuecomment-701543731)).

### DIFF
--- a/specs/proto/consensus.proto
+++ b/specs/proto/consensus.proto
@@ -4,16 +4,16 @@ syntax = "proto3";
 
 // https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#feeheader
 message FeeHeader {
-  uint64 baseRate = 1;
-  uint64 tipRate = 2;
+  uint64 base_rate = 1;
+  uint64 tip_rate = 2;
 }
 
 // https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#availabledataheader
 message AvailableDataHeader {
   // array of 32-byte hashes
-  repeated bytes rowRoots = 1;
+  repeated bytes row_roots = 1;
   // array of 32-byte hashes
-  repeated bytes colRoots = 2;
+  repeated bytes col_roots = 2;
 }
 
 message ConsensusProposal {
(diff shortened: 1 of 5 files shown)


specs/proto/consensus.proto
```diff
@@ -0,0 +1,54 @@
...
+  bytes last_header_hash = 4;
+  // 32-byte hash
+  bytes last_commit_hash = 5;
+  // 32-byte hash
+  bytes consensus_root = 6;
+  FeeHeader fee_header = 7;
+  // 32-byte hash
+  bytes state_commitment = 8;
+  uint64 available_data_original_shares_used = 9;
+  AvailableDataHeader available_data_header = 10;
```
liamsi, line 33: Didn't you say `available_data_original_shares_used` wasn't necessary as it can be inferred from the `available_data_header`?

adlerjohn: Nono, the _square size_ can be inferred from the available data header. How many shares of that square are used still needs to be provided.

liamsi: Ah that makes sense!

liamsi (approved):
LGTM 👍🏼

Issue #48, opened 2020-07-02
Spec Merkle multiproof format


adlerjohn:

liamsi:
Regarding the range proof format for the NMT, here is how I think they should look like:

```go
// Proof represents proof of a namespace.ID in an NMT.
// In case this proof proves the absence of a namespace.ID
// in a tree it also contains the leaf hashes of the range
// where that namespace would be.
type Proof struct {
	// Start index of this proof.
	Start int
	// End index of this proof.
	End int
	// Nodes that together with the corresponding leaf values
	// can be used to recompute the root and verify this proof.
	Nodes [][]byte
	// LeafHashes are nil if the namespace is present in the NMT.
	// In case the namespace to be proved is in the min/max range of
	// the tree but absent, this will contain the leaf hashes
	// necessary to verify the proof of absence.
	LeafHashes [][]byte
}
```
These probably should not be `int`s and for `nodes` as well as `leafHashes` we might want to use a type alias (or even a struct that captures `min`, `max`, and the actual digest).

musalbas:
I assume that `Nodes` are the subtree roots?

liamsi:
Yes, exactly. I think the name is appropriate because each subtree root is also a node in the tree, hence `proof.Nodes()`. `proof.SubtreeRoots()` would also be good.

liamsi:
> Note that to prove absence we need to provide the leafHashes too.

Don't we always just need a single leaf to prove absence? See:
https://github.com/lazyledger/nmt/pull/3#pullrequestreview-460049184

liamsi:
ref: https://github.com/lazyledger/lazyledger-core/issues/307#issuecomment-830673091

PR #129, opened 2021-02-19
Remove statement fields from Merkle proofs

### LINKED ISSUE #67: Proofs shouldn't include the Merkle root or the value being proven, as this is part of the statement being proven, not the proof
>A Merkle proof is a proof for the statement "a key x has value y for the tree with root z". It doesn't make sense for the statement being proven to be a part of the proof, because the prover could always generate a valid proof if they get to pick the statement. Furthermore, the verifier may already know what the statement being proven is, so including it as part of the proof may be redundant. At minimum, they should already know what the root of the tree is.

See: https://github.com/lazyledger/smt/pull/5/files#r460376202

adlerjohn:
Fixes #67. Does not address #48.

Removes "statement" (i.e. what you're trying to prove) fields from Merkle proof structs. Also clarify ordering of hashes. Leaf-to-root order was chosen to conform to https://tools.ietf.org/html/rfc6962#section-2.1.1.

### DIFF
--- a/specs/data_structures.md
+++ b/specs/data_structures.md
@@ -21,11 +21,11 @@ Data Structures
 - [Public-Key Cryptography](#public-key-cryptography)
 - [Merkle Trees](#merkle-trees)
   - [Binary Merkle Tree](#binary-merkle-tree)
-    - [BinaryMerkleTreeProof](#binarymerkletreeproof)
+    - [BinaryMerkleTreeInclusionProof](#binarymerkletreeinclusionproof)
   - [Namespace Merkle Tree](#namespace-merkle-tree)
-    - [NamespaceMerkleTreeProof](#namespacemerkletreeproof)
+    - [NamespaceMerkleTreeInclusionProof](#namespacemerkletreeinclusionproof)
   - [Sparse Merkle Tree](#sparse-merkle-tree)
-    - [SparseMerkleTreeProof](#sparsemerkletreeproof)
+    - [SparseMerkleTreeInclusionProof](#sparsemerkletreeinclusionproof)
 - [Erasure Coding](#erasure-coding)
   - [Reed-Solomon Erasure Coding](#reed-solomon-erasure-coding)
   - [2D Reed-Solomon Encoding Scheme](#2d-reed-solomon-encoding-scheme)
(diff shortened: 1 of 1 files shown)


liamsi (commented):
All for removing the "statements" from the proof. 

I'm also OK with explicitly stating the order of hashes but from @musalbas comment on discord he actually prefers to start from the root. 

Maybe break out the uncontroversial changes (removing the statements) to get them merged quicker?

musalbas:
Since the NebulousLabs library starts from the leaf, I'm ok with starting from the leaf and being consistent there for all Merkle tees.

liamsi (approved):
👍

specs/data_structures.md
```diff
@@ -346,18 +343,13 @@ For some intuition: the min and max namespace IDs for subtree roots with at leas
...
-| `key`                | `byte[32]`                       | Key (i.e. index) of the leaf. |
-| `siblingValues`      | [HashDigest](#hashdigest)`[]`    | Sibling hash values.          |
-| `siblingMins`        | [NamespaceID](#type-aliases)`[]` | Sibling min namespace IDs.    |
-| `siblingMaxes`       | [NamespaceID](#type-aliases)`[]` | Sibling max namespace IDs.    |
-| `leaf`               | `byte[]`                         | Leaf value.                   |
+| name            | type                             | description                                                     |
+| --------------- | -------------------------------- | --------------------------------------------------------------- |
+| `siblingValues` | [HashDigest](#hashdigest)`[]`    | Sibling hash values, ordered starting from the leaf's neighbor. |
+| `siblingMins`   | [NamespaceID](#type-aliases)`[]` | Sibling min namespace IDs.                                      |
+| `siblingMaxes`  | [NamespaceID](#type-aliases)`[]` | Sibling max namespace IDs.                                      |
```
liamsi, line 352: `NamespaceMerkleTreeInclusionProof` is the inclusion proof for a single node right? Isn't `siblingMins` and `Maxes` implicitly included in `siblingValues`?

adlerjohn: Yes, it's only for the inclusion of a single leaf. The min and max sibling values are committed to in the hashes, but they're not known to a verifier. So they still need to be provided.

PR #131, opened 2021-02-24
Add temporary linked list for paid shares

### LINKED ISSUE #88: Can't prove messages aren't paid for
One issue with the current scheme of separating transactions and messages is that while we can compactly prove that the message a transaction pays for has been included, we can't compactly prove that a message wasn't paid for.

The most obvious solutions are:
1. Add some reserved bytes at the beginning of each message pointing to the transaction index that pays for them. This doesn't work without major changes, as it breaks the non-interactive default share layout.
1. Accumulate <key: message start, value: <transaction index, message end (or message length)>> pair in a Sparse Merkle Tree and commit to the root in the block header. This way we can prove exclusion directly.
1. Require that transactions that pay for a message are ordered by message start (and implicitly namespace ID). This way we can prove exclusion with two inclusion proofs.

adlerjohn:
Fixes celestiaorg/celestia-specs#88. Follows the design outlined here https://github.com/lazyledger/lazyledger-specs/issues/88#issuecomment-785115580

### DIFF
--- a/specs/consensus.md
+++ b/specs/consensus.md
@@ -82,6 +82,7 @@ Consensus Rules
 | `ACTIVE_VALIDATORS_SUBTREE_ID`   | `StateSubtreeID` | `0x02` |
 | `INACTIVE_VALIDATORS_SUBTREE_ID` | `StateSubtreeID` | `0x03` |
 | `DELEGATIONS_SUBTREE_ID`         | `StateSubtreeID` | `0x04` |
+| `MESSAGE_PAID_SUBTREE_ID`        | `StateSubtreeID` | `0x05` |
 
 ### Rewards and Penalties
 
@@ -237,6 +238,25 @@ function validatorQueueRemove(validator, sender)
 
 Note that light clients cannot perform a linear search through a linked list, and are instead provided logarithmic proofs (e.g. in the case of `parentFromQueue`, a proof to the parent is provided, which should have `address` as its next validator).
 
+In addition, three helper functions to manage the [message paid list](./data_structures.md#messagepaid):
+1. `findFromMessagePaidList(start)`, which returns the transaction ID of the last transaction in the [message paid list](./data_structures.md#messagepaid) with `finish` greater than `start`, or `0` if the list is empty or no transactions in the list have at least `start` `finish`.
+1. `parentFromMessagePaidList(txid)`, which returns the transaction ID of the parent in the message paid list of the transaction with ID `txid`, or `0` if `txid` is not in the list or is the head of the list.
+3. `messagePaidListInsert`, defined as
+```py
+function messagePaidListInsert(tx, txid)
+    # Insert the new transaction into the linked list
+    parent = findFromMessagePaidList(tx.messageStartIndex)
+    state.messagesPaid[txid].start = tx.messageStartIndex
+    numShares = ceil(tx.messageSize / SHARE_SIZE)
+    state.messagesPaid[txid].finish = tx.messageStartIndex + numShares - 1
+    if parent != 0
+        state.messagesPaid[txid].next = state.messagesPaid[parent].next
+        state.messagesPaid[parent].next = txid
+    else
+        state.messagesPaid[txid].next = state.messagePaidHead
+        state.messagePaidHead = txid
+```
+
 We define a helper function to compute [F1 entries](../rationale/distributing_rewards.md):
 ```py
 function compute_new_entry(reward, power)
@@ -276,6 +296,8 @@ state.activeValidatorSet.proposerBlockReward += tipCost(bytesPaid)
 
 ```py
 bytesPaid = len(tx) + tx.messageSize
+currentStartFinish = state.messagesPaid[findFromMessagePaidList(tx.messageStartIndex)]
+parentStartFinish = state.messagesPaid[parentFromMessagePaidList(findFromMessagePaidList(tx.messageStartIndex))]
 ```
 
 The following checks must be `true`:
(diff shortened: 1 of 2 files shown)


specs/consensus.md
```diff
@@ -237,6 +238,25 @@ function validatorQueueRemove(validator, sender)
 
 Note that light clients cannot perform a linear search through a linked list, and are instead provided logarithmic proofs (e.g. in the case of `parentFromQueue`, a proof to the parent is provided, which should have `address` as its next validator).
 
+In addition, three helper functions to manage the [message paid list](./data_structures.md#messagepaid):
+1. `findFromMessagePaidList(start)`, which returns the transaction ID of the last transaction in the [message paid list](./data_structures.md#messagepaid) with `finish` greater than `start`, or `0` if the list is empty or no transactions in the list have at least `start` `finish`.
+1. `parentFromMessagePaidList(txid)`, which returns the transaction ID of the parent in the message paid list of the transaction with ID `txid`, or `0` if `txid` is not in the list or is the head of the list.
```
liamsi, line 243: Are txids defined somewhere?

specs/data_structures.md
```diff
@@ -920,6 +923,24 @@ For explanation on entries, see the [reward distribution rationale document](../
 
 Represents a (potentially) non-integer number.
 
+### MessagePaid
+
+| name     | type                      | description                                                           |
+| -------- | ------------------------- | --------------------------------------------------------------------- |
+| `start`  | `uint64`                  | Share index (in row-major order) of first share paid for (inclusive). |
+| `finish` | `uint64`                  | Share index (in row-major order) of last share paid for (inclusive).  |
```
liamsi, line 931: Nit: of first/last share paid for in this block? Or by a given set of Txs?

specs/consensus.md
```diff
@@ -237,6 +238,25 @@ function validatorQueueRemove(validator, sender)
...
+    parent = findFromMessagePaidList(tx.messageStartIndex)
+    state.messagesPaid[txid].start = tx.messageStartIndex
+    numShares = ceil(tx.messageSize / SHARE_SIZE)
+    state.messagesPaid[txid].finish = tx.messageStartIndex + numShares - 1
+    if parent != 0
+        state.messagesPaid[txid].next = state.messagesPaid[parent].next
+        state.messagesPaid[parent].next = txid
+    else
+        state.messagesPaid[txid].next = state.messagePaidHead
+        state.messagePaidHead = txid
```
liamsi, line 257: Something I've missed last time: should we make these helpers pure functions (i.e. pass in a `state` and return a state)? (only a suggestion and orthogonal to this PR as it would touch all pseudo code)

adlerjohn: Ehhh they operator on state, so maybe making them methods of a state class would make more sense? Which would make them implicitly pure functions on an implicit `self` pointer.

specs/consensus.md
adlerjohn: Ugh no, but that's an orthogonal issue and something that needs to be specified too. I'll open an issue.

specs/data_structures.md
adlerjohn: Neither? For the transaction with the ID that's the same as the key of this state element.

specs/consensus.md
liamsi: I'm in favor of not using anything fancier than functions (no classes / OOP paradigms).

specs/data_structures.md
liamsi: I see.

specs/consensus.md
liamsi: Reason being that they are easier to grasp and read (I find OOP python ugly), and afaik easier to formally verify (the latter could be the most interesting reason for a spec). Does the ethereum spec contain classes btw?

adlerjohn: We'll see once I get to writing the stuff in python. The Serenity spec passes in a state pointer explicitly.

liamsi: And I can see classes: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/lightclient/beacon-chain.md#beaconstate
>

liamsi: So, is that state class defined somewhere then?

adlerjohn: I mean yeah, the alternative to classes would be a bunch of unorganized globals! Here's an example of passing in a pointer to the state https://github.com/ethereum/eth2.0-specs/blob/dev/specs/lightclient/beacon-chain.md#get_sync_committee_indices . Though that may just be how python does classes.

adlerjohn: Our state "class" is kind of defined here https://github.com/lazyledger/lazyledger-specs/blob/0d5ad1f446accd1798389bedfbd31b4a4e5f1794/specs/consensus.md#state-transitions

liamsi: >  Though that may just be how python does classes.

Is that really runnable code? I like structuring it in state accessors (which are indeed pure functions) and state mutators.

liamsi: > Our state "class" is kind of defined here https://github.com/lazyledger/lazyledger-specs/blob/0d5ad1f446accd1798389bedfbd31b4a4e5f1794/specs/consensus.md#state-transitions

That makes sense but I think the methods should take in a state parameter at least.

liamsi (approved):
LGTM

metatron1973:
When do you place Bug label ?

metatron1973 (commented):
What is fastest way to detect any Bugs?

PR #144, opened 2021-03-16
Simplify DA header to commit to full row and column roots

### LINKED ISSUE #140: Simplify DA header commitment
[Currently](https://github.com/lazyledger/lazyledger-specs/blob/b3c4122b2503bebd3bdd46d6e6f1a570de5f5c0f/specs/data_structures.md#availabledataheader), the available data header contains all column roots and a subset of row roots. Rows with only tail padding are available and their roots can be computed locally, so they are omitted.

Pro of this approach: it reduces the size of the DA header by up to 25% (up to half of the original square rows).

Con of this approach: it introduces more complexity in the calculation of the Merkle root of row and column roots, since the number of row and column roots is not equal.

It turns out reducing the size of the DA header can be done entirely implementation-side, logically through placeholder values in a lookup table. As an example, an implementation that receives as input 9 row roots and 16 column roots knows that there are 7 rows of tail padding, and can insert the appropriate roots from a lookup table.

The computation of the available data root in the block header can be simplified to always use the same number of row and column roots, and defer optimization to the implementation (i.e. make optimizations optional, thus simplifying the spec).

adlerjohn:
Fixes #140.

### DIFF
--- a/specs/consensus.md
+++ b/specs/consensus.md
@@ -140,7 +140,7 @@ The [block header](./data_structures.md#header) `block.header` (`header` for sho
 
 The [available data header](./data_structures.md#availabledataheader)) `block.availableDataHeader` (`availableDataHeader` for short) is then processed. This commits to the available data, which is only downloaded after the [consensus commit](#blocklastcommit) is processed. The following checks must be `true`:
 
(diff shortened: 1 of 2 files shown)


musalbas (approved):

liamsi (approved):
👍🏼

PR #146, opened 2021-03-17
Erasure-code namespace ID

### LINKED ISSUE #145: Namespace ID must be erasure-coded
Currently, what is erasure-coded does not include the namespace ID. Therefore, in the event a row of column needs to be recovered, the namespace information cannot also be recovered. The namespace ID for each share should be part of the leaf for NMTs.

Fix: `node.v = h(0x00, d.rawData)` to `node.v = h(0x00, d.namespaceID, d.rawData)`

(Thanks @evan-forbes for noticing this.)

adlerjohn:
Fixes #145.

Note the proposed fix in #145 is actually not complete as it does not affect erasure coding, but rather only the NMT. This PR changes the format of non-parity shares themselves to be 8-byte namespace ID + 248 bytes of data, which is then erasure coded. This guarantees a power-of-2 invariant on share size, which is needed for proper erasure coding.


specs/data_structures.md
```diff
@@ -330,7 +330,7 @@ For leaf node `node` of [share](#share) data `d`:
 ```C++
 node.n_min = d.namespaceID
 node.n_max = d.namespaceID
-node.v = h(0x00, d.rawData)
+node.v = h(0x00, d.namespaceID, d.rawData)
```
liamsi, line 333: I don't know how often we've already talked about this line :-D

adlerjohn: Yeah I'm sure we even discussed adding in the namespace ID here, but not in the context of the recovering case.

adlerjohn:
Converting to draft while we figure out the full extent of changes required.

specs/data_structures.md
liamsi, line 333: So this would only be the case for original data. For parity shares, this would still stay:
`node.v = h(0x00, d.rawData)`
right? @evan-forbes @adlerjohn

adlerjohn: This would actually not be the case of original data either! (Pushing the fix soon)

liamsi:
render version of the the share section: https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn-nmt_namespace_leaves/specs/data_structures.md#share

specs/data_structures.md
```diff
@@ -496,13 +496,15 @@ If a malicious block producer incorrectly computes the 2D Reed-Solomon code for
...
 
 ![fig: Reserved share.](./figures/share.svg)
 
-For shares **with a namespace ID above [`NAMESPACE_ID_MAX_RESERVED`](./consensus.md#constants)**, the first [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes have no special meaning and are simply used to store data like all the other bytes in the share.
+For shares **with a namespace ID above [`NAMESPACE_ID_MAX_RESERVED`](./consensus.md#constants)**, the first [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes after [`NAMESPACE_ID_BYTES`](./consensus.md#constants) have no special meaning and are simply used to store data like all the other bytes in the share.
```
liamsi, line 508: This whole paragraph in its current form rendered incl. the vector graphic:
image

liamsi: So because of this `*` the raw data can only be max 247 = 256 - 1 - 8 bytes long right?

liamsi: Ah I see, the next section clarifies this:
`SHARE_SIZE - NAMESPACE_ID_BYTES - SHARE_RESERVED_BYTES`

specs/data_structures.md
```diff
@@ -516,7 +518,7 @@ Then,
...
         1. Compute the length of each serialized request, [serialize the length](#share), and pre-pend the serialized request with its serialized length.
-    1. Split up the length/request pairs into [`SHARE_SIZE`](./consensus.md#constants)`-`[`SHARE_RESERVED_BYTES`](./consensus.md#constants)-byte [shares](#share) and assign [the appropriate namespace ID](./consensus.md#reserved-namespace-ids). This data has a _reserved_ namespace ID, so the first [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes for these shares must be [set specially](#share).
+    1. Split up the length/request pairs into [`SHARE_SIZE`](./consensus.md#constants)`-`[`NAMESPACE_ID_BYTES`](./consensus.md#constants)`-`[`SHARE_RESERVED_BYTES`](./consensus.md#constants)-byte [shares](#share) and assign [the appropriate namespace ID](./consensus.md#reserved-namespace-ids). This data has a _reserved_ namespace ID, so the first [`NAMESPACE_ID_BYTES`](./consensus.md#constants)`+`[`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes for these shares must be [set specially](#share).
 1. Concatenate the lists of shares in the order: transactions, intermediate state roots, evidence.
```
liamsi, line 522: So the concatenation is what makes it necessary to use the `*` thingy (the start index above)? 
It feels the text is missing some step and does not match the diagram: the concatenation here can refer to what leads to 
image
because if you simply concatenate, tx3 would also carry a namespace.
Other than this concatenation, the text only mentions to split up requests into share_size - 8 - 1 shares.

specs/data_structures.md
```diff
@@ -496,13 +496,15 @@ If a malicious block producer incorrectly computes the 2D Reed-Solomon code for
...
 
 ![fig: Reserved share.](./figures/share.svg)
 
-For shares **with a namespace ID above [`NAMESPACE_ID_MAX_RESERVED`](./consensus.md#constants)**, the first [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes have no special meaning and are simply used to store data like all the other bytes in the share.
+For shares **with a namespace ID above [`NAMESPACE_ID_MAX_RESERVED`](./consensus.md#constants)**, the first [`SHARE_RESERVED_BYTES`](./consensus.md#constants) bytes after [`NAMESPACE_ID_BYTES`](./consensus.md#constants) have no special meaning and are simply used to store data like all the other bytes in the share.
 
 For non-parity shares, if there is insufficient request data to fill the share, the remaining bytes are padded with `0`.
```
liamsi, line 509: > insufficient request data

What does that mean? As far as I can see from the diagram, if there is still space left in the share, remaining Tx are squeezed into the share. So this could only happen in the very last share. Is that right?

liamsi (commented):
TBH, this is still ambiguous. I think I understand how this works but I asked a lot of questions to better be sure.

specs/data_structures.md
liamsi: Also, with the above description (step 1), it is not clear if there shouldn't be an associated NID for the the length/request pair len(tx3), tx3.

specs/data_structures.md
adlerjohn: Yes, but also for messages. Non-parity shares are for both transactions and messages.

specs/data_structures.md
adlerjohn: Uhh okay I see why the text might be confusing. What it's supposed to be is that the namespace ID of all transactions is the same, so only shares have a namespace ID, not individual transactions.

adlerjohn: Can clarify.

specs/data_structures.md
liamsi: Maybe the encoding is wrong in the implementation but note that the index, if varint encoded could exceed one byte: https://github.com/lazyledger/lazyledger-app/issues/53#issuecomment-802638066

adlerjohn: Fixed to 1 byte in 5b010a7 as per https://github.com/lazyledger/lazyledger-app/issues/53#issuecomment-802811566

specs/data_structures.md
adlerjohn: Clarified I think in 42d8b4f.

liamsi (approved):
LGTM, thanks for the clarifications.

PR #151, opened 2021-03-30
Clarify minimum square size = 1

adlerjohn:
As per https://github.com/lazyledger/lazyledger-core/pull/248#discussion_r602877505 clarifies the edge case of minimum square size.


liamsi (approved):
👍

Issue #154, opened 2021-04-29
Remove or at least revisit redundant fields


liamsi:
## Context
The (consensus message) Proposal adds a few fields to the spec that do not exist in tendermint. It is unclear why this new fields are even needed and I think they do not really add any additional benefit.

Proposal here:

https://github.com/lazyledger/lazyledger-specs/blob/de5f4f74f56922e9fa735ef79d9e6e6492a2bad1/specs/proto/consensus.proto#L19-L36

Proposal in tendermint:

https://github.com/tendermint/tendermint/blob/8eccaf95358124b595659b55c465ff2228c7168b/types/proposal.go#L19-L33

And CanonicalProposal in tendermint (this is what actually gets signed by validators):
https://github.com/tendermint/tendermint/blob/8eccaf95358124b595659b55c465ff2228c7168b/proto/tendermint/types/canonical.pb.go#L135-L143

See this discussion in https://github.com/lazyledger/lazyledger-core/pull/305: 
https://github.com/lazyledger/lazyledger-core/pull/305#discussion_r622984581

## Suggestion

Remove all fields that are not absolutely required to efficiently do DA sampling.

Issue #153, opened 2021-04-28
POLRound missing from Proposal


liamsi:
We can't simply remove the POLRound as it is essential for the tendermint consensus to function properly.

From the tendermint specs:

- [Proposal](https://github.com/tendermint/spec/blob/84ee4249aea6740a1149cc8ca47b6e9f123305c1/spec/core/data_structures.md#proposal)
- [ProposalPOL ](https://github.com/tendermint/spec/blob/72d15a4b0773adb8d5160a34d9a8e2d981394484/spec/p2p/messages/consensus.md#proposalpol)

Also see:
https://arxiv.org/pdf/1807.04938.pdf explains the tendermint algorithm in its current form (Algorithm 1) but it is actually non-trivial to see how this directly relates to POLRound field. 

This contains a simple description of the polka: https://github.com/tendermint/tendermint/blob/master/docs/introduction/what-is-tendermint.md#consensus-overview

See also: https://github.com/lazyledger/lazyledger-core/blob/1a08b430a8885654b6e020ac588b1080e999170c/types/proposal.go#L22-L25

PR #160, opened 2021-05-10
Align specs with base Tendermint and current lazyledger-core impl

### LINKED ISSUE #139: specs -> impl
## Summary

We should open issues in lazyledger-core with the changes to core types. 

This brings light to the required changes and gets people thinking about the implementation details. 

We can add a short sentence or paragraph to the contributing.md in order for people to know to do this.

adlerjohn:
Fixes #139. (Explicitly does not address #153 or #154).

Aligns specs with the current data structures (and implied consensus rules) in https://github.com/lazyledger/lazyledger-core.


rationale/fees.md
```diff
@@ -6,6 +6,8 @@
 
 ## Preamble
 
+**This document will apply to a future version of the LazyLedger protocol.**
```
liamsi, line 9: What we could do: we could tag the current state of the spec (`v0.1.0-idealized`) and remove all things that we plan to add later. That might be clearer than this and we have a tagged version that we can refer to in future discussions?

liamsi: > (it can always be referenced by commit hash).

Sure, but I feel like adding a tag is easier to find and more explicit.

specs/consensus.md
```diff
@@ -44,23 +44,24 @@
...
+| `STATE_SUBTREE_RESERVED_BYTES`          | `uint64` | `1`            | `byte`  | Number of bytes reserved to identify state subtrees.                                                                                                                |
+| `UNBONDING_DURATION`                    | `uint32` |                | `block` | Duration, in blocks, for unbonding a validator or delegation.                                                                                                       |
+| `VERSION_APP`                           | `uint64` | `1`            |         | Version of the LazyLedger application. Breaking changes (hard forks) must update this parameter.                                                                    |
+| `VERSION_BLOCK`                         | `uint64` | `1`            |         | Version of the LazyLedger chain. Breaking changes (hard forks) must update this parameter.                                                                          |
```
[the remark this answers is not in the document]
liamsi: Wow, why are you doing this me github! WHY!!!!

liamsi:
> Fixes #139. (Explicitly does not address #153 or #154).

Not really about #139 though right? see:

> We should open issues in lazyledger-core with the changes to core types.

So the changes I see:
 - height, round -> int instead of uint
 - use Timestamp instead
 - remove FeeHeader and rationale doc about fees
 - Version (single int) -> ConsensusVersion struct
 
Is there anything else?

adlerjohn:
> Or are you saying that there are no changes left that need to be done to core types? I think there still are a few places in ll-core.

It's more that, with this PR, there are no more "big" changes that need to be made (e.g. that require changes to ABCI), or superfluous changes that aren't critical for functionality. The (very few remaining) changes that do need to be made are in the process of being taken care of in the ll-core repo, so #139 is no longer really relevant.

> Is there anything else?

There's also
1. Re-adding version and chain ID to the header.
2. Changing consensus root to consensus hash.

specs/data_structures.md
```diff
@@ -79,19 +79,19 @@
...
-| `StateSubtreeID`            | `byte`                     |
-| [`Timestamp`](#timestamp)   | `uint64`                   |
-| `VotingPower`               | `uint64`                   |
+| name                        | type                        |
+|-----------------------------|-----------------------------|
+| [`Address`](#address)       | `byte[32]`                  |
+| `Amount`                    | `uint64`                    |
+| `Graffiti`                  | `byte[MAX_GRAFFITI_BYTES]`  |
+| [`HashDigest`](#hashdigest) | `byte[32]`                  |
+| `Height`                    | `int64`                     |
```
Wondertan, line 88: `int64` vs `uint64`? Idk exactly

If there is no difference, changing the specs to conform impl is easier. Otherwise, I can change the impl, like I did in [#315]https://github.com/lazyledger/lazyledger-core/pull/315() also if uin64 is in some way better.  
Same for `Round`

specs/proto/consensus.proto
```diff
@@ -17,22 +19,21 @@ message AvailableDataHeader {
...
   // 32-byte hash
-  bytes consensus_root = 6;
-  FeeHeader fee_header = 7;
+  bytes consensus_hash = 6;
   // 32-byte hash
-  bytes state_commitment = 8;
-  uint64 available_data_original_shares_used = 9;
-  AvailableDataHeader available_data_header = 10;
+  bytes state_commitment = 7;
+  uint64 available_data_original_shares_used = 8;
```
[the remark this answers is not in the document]
Wondertan: Maybe then a new issue is needed for integer consistency?

PR #161, opened 2021-05-10
Fix min row/col roots to 2

adlerjohn:
Since the minimum ODS is 1x1, the minimum EDS should be 2x2.


evan-forbes (approved):
makes sense. ref [#312](https://github.com/lazyledger/lazyledger-core/pull/312) which enforces this beyond what was done in [#282](https://github.com/lazyledger/lazyledger-core/pull/282)

PR #162, opened 2021-05-17
Add Proof of Lock round to proposal struct

### LINKED ISSUE #153: POLRound missing from Proposal

adlerjohn:
Fixes #153 and #154.

This adds the `POLRound` field to consensus proposals. It also refactors [proposals](https://github.com/lazyledger/lazyledger-core/blob/66dd73241e508bc439679afd49683282f8d47b5c/proto/tendermint/types/types.proto#L188-L198) and [votes](https://github.com/lazyledger/lazyledger-core/blob/66dd73241e508bc439679afd49683282f8d47b5c/proto/tendermint/types/types.proto#L155-L168) to conform to the implementation.

~~One difference between the implementation and this PR is the removal of the `available_data_header` field from proposals: https://github.com/lazyledger/lazyledger-specs/compare/adlerjohn/polround?expand=1#diff-48a78bd7e15d8c7211897da188db99d6576a27e125cbde564ae0238ae66e9e94L34~~

~~While the implementation [currently](https://github.com/lazyledger/lazyledger-core/blob/66dd73241e508bc439679afd49683282f8d47b5c/types/proposal.go#L33) has this field, that shouldn't be the case. The rest of the block header is distributed separately from the proposal, which [only commits to the block header](https://github.com/lazyledger/lazyledger-core/blob/66dd73241e508bc439679afd49683282f8d47b5c/types/proposal.go#L30). This is being changed in https://github.com/lazyledger/lazyledger-core/pull/312.~~

### DIFF
--- a/specs/proto/consensus.proto
+++ b/specs/proto/consensus.proto
@@ -16,6 +16,14 @@ enum SignedMsgType {
   SignedMsgTypeProposal = 32;
 }
 
+// https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#availabledataheader
+message AvailableDataHeader {
+  // array of 32-byte hashes
+  repeated bytes row_roots = 1;
+  // array of 32-byte hashes
+  repeated bytes col_roots = 2;
+}
+
 message ConsensusProposal {
   SignedMsgType type = 1;
   int64 height = 2;
@@ -27,6 +35,7 @@ message ConsensusProposal {
   google.protobuf.Timestamp timestamp = 6;
   // 64-byte signature
   bytes proposer_signature = 7;
+  AvailableDataHeader da_header = 8;
 }
 
 message ConsensusVote {
--- a/specs/proto/consensus.proto
+++ b/specs/proto/consensus.proto
@@ -36,7 +36,9 @@ message ConsensusVote {
   // 32-byte hash
   bytes header_hash = 4;
   google.protobuf.Timestamp timestamp = 5;
+  // 32-byte hash
   bytes validator_address = 6;
+  int32 validator_index = 7;
   // 64-byte signature
-  bytes signature = 7;
+  bytes signature = 8;
 }
--- a/specs/proto/consensus.proto
+++ b/specs/proto/consensus.proto
@@ -4,52 +4,39 @@ import "google/protobuf/timestamp.proto";
 
 // Define wire types for consensus messages.
 
-// https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#consensusversion
-message ConsensusVersion {
-  uint64 block = 1;
-  uint64 app = 2;
-}
+// SignedMsgType is a type of signed message in the consensus.
+enum SignedMsgType {
+  SignedMsgTypeUnknown = 0;
 
-// https://github.com/lazyledger/lazyledger-specs/blob/master/specs/data_structures.md#availabledataheader
-message AvailableDataHeader {
-  // array of 32-byte hashes
-  repeated bytes row_roots = 1;
-  // array of 32-byte hashes
-  repeated bytes col_roots = 2;
+  // Votes
+  SignedMsgTypePrevote = 1;
+  SignedMsgTypePrecommit = 2;
+
+  // Proposals
+  SignedMsgTypeProposal = 32;
 }
 
 message ConsensusProposal {
-  int64 height = 1;
-  int32 round = 2;
-  google.protobuf.Timestamp timestamp = 3;
+  SignedMsgType type = 1;
+  int64 height = 2;
+  int32 round = 3;
+  int32 pol_round = 4;
   // 32-byte hash
-  bytes last_header_hash = 4;
-  // 32-byte hash
-  bytes last_commit_hash = 5;
-  // 32-byte hash
-  bytes consensus_hash = 6;
-  // 32-byte hash
-  bytes state_commitment = 7;
-  uint64 available_data_original_shares_used = 8;
-  AvailableDataHeader available_data_header = 9;
+  // Proposed block header
+  bytes header_hash = 5;
+  google.protobuf.Timestamp timestamp = 6;
   // 64-byte signature
-  bytes proposer_signature = 10;
-}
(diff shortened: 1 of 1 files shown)


liamsi:
> One difference between the implementation and this PR is the removal of the available_data_header field from proposals:
> While the implementation currently has this field, that shouldn't be the case. The rest of the block header is distributed separately from the proposal, which only commits to the block header.

Wait, I'm pretty sure we had a discussion to include the DA header in the proposal as anyone who cares about the proposal will need it anyways and we decided this is an optimization we want to save some roundtrips... I can dig up up the discussions later.

Technically, at least the data hash should be included in the proposal (the header hash is more than that).

adlerjohn:
> Wait, I'm pretty sure we had a discussion to include the DA header in the proposal as anyone who cares about the proposal will need it anyways and we decided this is an optimization we want to save some roundtrips...

You still need an extra request because you have to download the header. The optimization was originally in the context of the proposal including all header fields (i.e. as the proposal is before this PR). There's two alternatives here if you want to avoid an _additional_ request to download the DA header.
1. Include the DA header with the proposal.
1. Include the DA header in a new struct, e.g. HeaderPlusDAHeader, that includes both the header and DA header.

Both work, (2) is cleaner. But if you're willing to break the proposal by adding the DA header...why not just add the whole header to the proposal? Then you can save the second request to download the header.

liamsi:
I need to think about this in the context of the implementation. I'll reserve some time for this towards the end of the week. I'm reluctant to merge this as is because it implies that validators receiving the proposal will have to download the header. Currently, this is implicitly done via the gossip routine that "pushes" out Parts (the Parts include the whole block, including the header - not just the Block.Data). We will have to think about this more concretely, i.e. at a protocol level, not just from the data structures. Will the other validators receive the header and all relevant data to make a decision on a proposal, or, will they actively download the header and how resepectively.

liamsi:
> The optimization was originally in the context of the proposal including all header fields (i.e. as the proposal is before this PR). 

I think what is confusing to me is that in previous conversations (https://github.com/lazyledger/lazyledger-core/pull/305#discussion_r622984581) it was mentioned that these fields are completely redundant. Now it sounds they are necessary and not redundant because without them, (das) validators won't be able to reconstruct the header?

liamsi:
Also, the question is: which values do the validators sign on proposal? Previously, that was the BlockID which as mentioned above was a representation of the full block (including the header). In our case, will it be the header hash or the data hash? I think it should be the header has, as the data hash does only commit to block data and not to height, time etc.

Would also like to loop in @Wondertan who did start a consensus reactor ADR. Would be curious to know how that aligns with your thoughts.

specs/proto/consensus.proto
```diff
@@ -4,52 +4,41 @@ import "google/protobuf/timestamp.proto";
...
+  SignedMsgType type = 1;
+  int64 height = 2;
+  int32 round = 3;
+  int32 pol_round = 4;
   // 32-byte hash
-  bytes state_commitment = 7;
-  uint64 available_data_original_shares_used = 8;
-  AvailableDataHeader available_data_header = 9;
+  // Proposed block header
+  bytes header_hash = 5;
```
liamsi, line 26: So if a proposal only includes the header hash, is the assumption that validators will _receive_ the header (and the DAHeader) via gossiping (if yes, why not include everything they need for validating the block in the proposal instead?), or, will they have to _request_ it (if so by whom and how).

adlerjohn: Tendermint and lazyledger-core currently only include the BlockID (which is the header hash + parts). Since we're removing parts, essentially they only include the header hash, not the fully header. Therefore, they must receive the block header separately _currently_.

The current specs (prior to this PR) optimize this by include all header information (including the DA header) into the proposal:

https://github.com/lazyledger/lazyledger-specs/blob/bfd7d0019a5d4f3f00dbbb263818b47be8bbb155/specs/proto/consensus.proto#L21-L37

This PR regresses that optimization to be in line with the implementation, and also prescribes regressing the optimization of including the DA header in the proposal since it's inconsistent with the exclusion of the header.

liamsi: So these fields never were redundant, right?

liamsi:
> But if you're willing to break the proposal by adding the DA header...

We must break the proposal either way (besides leaving consensus protocol details in place, like POLRound). 

The question rather is: how exactly and _why_? What are the assumptions you are making about the protocol that uses the Proposal? I understand you are assuming that validators will a) download (or retrieve?) the header and b) download (or retrieve?) the DAHeader? I think instead of just adding or removing fields we have to clearly specify how they will be used in this PR (or at least clearly specifying the assumptions made).

specs/proto/consensus.proto
adlerjohn: They are still redundant. If a new proposal is received from an honest node, most of those values can be inferred from the current state. E.g. height must be previous height + 1. The last header hash is the previous header's hash. And so on.

liamsi: How would a DAS node perviously e.g. infer `state_commitment`, `available_data_original_shares_used`?

liamsi (approved):

Issue #94, opened 2020-11-18
Add implicit and canonical data fields


adlerjohn:
E.g. transaction signatures should sign over the chain ID to prevent replay attacks and other exploits, but including the chain ID for each transaction is redundant and can instead be implicit. Same for votes.

adlerjohn:
cosmos sdk

PR #164, opened 2021-05-22
Move header and DA header into proposal

### LINKED ISSUE #163: Consider moving header and DA header into proposal
Ref #162.

adlerjohn:
Fixes #163.

This optimization reduces the number of network round trips by putting both the header and DA header directly into the proposal, so that a single network message is needed to get everything except the block body.

Note that canonicalization is explicitly not addressed here, but will be in #94. Properly abstracted, canonicalization should only affect signature verification and initial network handshake, so _should_ be unimportant for processing.

### DIFF
--- a/specs/networking.md
+++ b/specs/networking.md
@@ -24,6 +24,31 @@
 
 Defined as `ConsensusProposal` [here](./proto/consensus.proto).
 
+When receiving a new block proposal `proposal` from the network, the following steps are performed in order. _Must_ indicates that peers must be blacklisted (to prevent DoS attacks) and _should_ indicates that the network message can simply be ignored.
+
+1. `proposal.type` must be a `SignedMsgType`.
+1. `proposal.round` is processed identically to Tendermint.
+1. `proposal.pol_round` is processed identically to Tendermint.
+1. `proposal.header` must be well-formed.
+1. `proposal.header.version.block` must be [`VERSION_BLOCK`](./consensus.md#constants).
+1. `proposal.header.version.app` must be [`VERSION_APP`](./consensus.md#constants).
+1. `proposal.header.height` should be previous known height + 1.
+1. `proposal.header.chain_id` must be [`CHAIN_ID`](./consensus.md#constants).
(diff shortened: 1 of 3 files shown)


specs/networking.md
```diff
@@ -19,6 +20,35 @@
...
+1. `proposal.header.available_data_original_shares_used` must be at most [`AVAILABLE_DATA_ORIGINAL_SQUARE_MAX ** 2`](./consensus.md#constants).
+1. `proposal.header.available_data_root` must be the [root](./data_structures.md#availabledataheader) of `proposal.da_header`.
+1. `proposal.header.proposer_address` must be the [correct leader](./consensus.md#leader-selection).
+1. `proposal.da_header` must be well-formed.
+1. The number of elements in `proposal.da_header.row_roots` and `proposal.da_header.row_roots` must be equal.
+1. The number of elements in `proposal.da_header.row_roots` must be the same as computed [here](./data_structures.md#header).
+1. `proposal.proposer_signature` must be a valid [digital signature](./data_structures.md#public-key-cryptography) over the header hash of `proposal.header` that recovers to `proposal.header.proposer_address`.
+1. For [full nodes](./node_types.md#node-type-definitions), `proposal.da_header` must be the result of computing the roots of the shares (received separately).
+1. For [light nodes](./node_types.md#node-type-definitions), `proposal.da_header` should be sampled from for availability.
```
liamsi, line 50: This is dope!

In general, I'm in favor of a) stating there is a validity-predicate isValidProposal(...) and b) a pseudo-code listing that clarifies input of that predicate and lists validation steps in pseude-code instead.

I can provide that as a follow-up to this PR.

adlerjohn: I agree all those are good things. Would prefer deferring to a future PR though as essentially a proto-ADR (or an actual ADR?).

liamsi: Yeah, we can do it in followup PRs. Just noting that pseudo-code would be better independent of existing implementation infrastructure etc.

liamsi (approved):
Thanks for this 👍🏼

PR #167, opened 2021-05-29
Add CI action to deploy book to GitHub pages

adlerjohn:
Follow up on #165. This action will automatically deploy `master` to `https://lazyledger.github.io/lazyledger-specs/latest` and tags to `https://lazyledger.github.io/lazyledger-specs/<tag>`. (Unfortunately no good way to preview the effects of this action until it runs.)

PR #168, opened 2021-05-29
Fix pages deploy action

adlerjohn:
Fix pages action to use `.` instead of `docs` 😬

https://github.com/lazyledger/lazyledger-specs/pull/167#issuecomment-850887306

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -18,13 +18,13 @@ jobs:
         with:
           mdbook-version: '0.4.8'
 
-      - run: mdbook build docs
+      - run: mdbook build .
 
       - name: Deploy latest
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -18,13 +18,13 @@ jobs:
         with:
           mdbook-version: '0.4.8'
 
-      - run: mdbook build docs
+      - run: mdbook build .
```
liamsi, line 21: hehe

liamsi (approved):

.github/workflows/gh-pages.yml
adlerjohn: 😬

Issue #175, opened 2021-05-31
node types: storage nodes


liamsi:
Partial nodes map to storage nodes in the original paper. IMO, the specification should mention this fact but independently of the original paper the specification should also mention the role of storage nodes for the p2p network.

PR #177, opened 2021-06-03
Note that partial nodes can act as storage nodes for subset of blocks

adlerjohn:
Ref: #175.

Note that storage nodes were already defined at the bottom: https://github.com/lazyledger/lazyledger-specs/blob/de7abfcbed571d5980b25c9ea2b182d141270232/src/specs/node_types.md#node-type-definitions

> Storage nodes provide the same security guarantees as full nodes. Block bodies (in erasure-coded form) are stored and served to the network. 

Expand node type descriptions.

Rendered: https://github.com/lazyledger/lazyledger-specs/blob/adlerjohn/expand_storage_nodes/src/specs/node_types.md#node-type-definitions

### DIFF
--- a/src/specs/node_types.md
+++ b/src/specs/node_types.md
@@ -79,29 +79,57 @@ At most secure under a weak subjectivity assumption.
 
 For convenience, we will define several common parameter configurations:
 
-1. [Full nodes](https://en.bitcoin.it/wiki/Full_node) provide the strongest security guarantees. Block bodies do not need to be stored.
+1. [Full nodes](https://en.bitcoin.it/wiki/Full_node) provide the strongest security guarantees.
     - Block headers: [Extended Block Headers](#extended-block-headers)
     - Block bodies: [Full Bodies](#full-bodies)
+        - Storage: Not stored beyond a parameter
+        - Serving: Served to the network
     - Transactions: [Full Transactions](#full-transactions)
+    - Fraud proofs:
+        - State transition: produce, consume
+        - Erasure coding: produce, consume
 1. Partial nodes are capable of producing fraud proofs of invalid transactions and contribute to validating the erasure coding of random blocks.
     - Block headers: [Extended Block Headers](#extended-block-headers)
     - Block bodies: [Partial Bodies](#partial-bodies)
+        - Storage: Not stored beyond a parameter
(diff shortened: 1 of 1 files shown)


src/specs/node_types.md
```diff
@@ -103,3 +103,5 @@ For convenience, we will define several common parameter configurations:
     - Block headers: [Extended Block Headers](#extended-block-headers)
     - Block bodies: [Full Bodies](#full-bodies)
     - Transactions: [Full Transactions](#full-transactions)
+
+Note: partial nodes may also serve the purpose of storage nodes (distributing shares to the network) for a subset of blocks.
```
liamsi, line 107: Oh, I see. Hmm, the nomenclature is kinda confusing.

So the difference between a Storage Node and a Full Node is only the fact that one stores the erasure-coded Block bodies and the other doesn't necessarily (e.g. could be validators). Otherwise, they are the same. For partial nodes it remains unclear if they store (parts of the block bodies) or not until this Note at the bottom.

Here is what I think could improve this:

Remove the specific node-type. Either add a bullet-point to each other called "Stores:" and list what it is expected to store (and if it is optional). Or, add a section about storage as all other nodes types (besides supe-light-nodes) will in reality  "store and served to the network".

While at it:
> Light validator nodes can produce new blocks with strong security guarantees and light resource requirements.

We decided that this node-type does not exist anymore.

liamsi (changes requested):
> Light validator nodes can produce new blocks with strong security guarantees and light resource requirements.

While this is slightly orthogonal, I would strongly suggest removing this type. We decided last week to drop this node type.

Left some opinionated suggestions how to improve the other changes: https://github.com/lazyledger/lazyledger-specs/pull/177#discussion_r645121588

IMO, this still needs to be clearer.

liamsi:
Also wanted to bring in the perspective that the node-types section specification should inform this: https://github.com/lazyledger/lazyledger-core/issues/388 and will play a big role in defining parts of next big milestone: https://github.com/lazyledger/lazyledger-core/issues/381

liamsi (approved):
I feel there is still some redundancy between the definitions between the different node types and this could be further simplified (i.e. this should be a table).

But this is already so much better!

PR #184, opened 2021-06-27
Add no-DAS validator node type

### LINKED ISSUE #180: Codify the decision that validators will download all block data
On various chats and synchronous calls, it was decided that validators will not perform data availability sampling, and will instead download all of the block data. I couldn't find this decision written down formally anywhere and figured that here is the place to do it.

While I couldn't find any, we might also need to change portions of the spec that rely on the assumption that validators will not be downloading all of the block data.

adlerjohn:
Fixes celestiaorg/celestia-specs#180

### DIFF
--- a/src/specs/node_types.md
+++ b/src/specs/node_types.md
@@ -133,3 +133,12 @@ For convenience, we will define several common parameter configurations:
     - Fraud proofs:
         - State transition: produce, consume
         - Erasure coding: produce (partial), consume
+1. Validator nodes are identical to full nodes, but can additionally created consensus messages (proposals, votes, etc.).
+    - Block headers: [Extended Block Headers](#extended-block-headers)
+    - Block bodies: [Full Bodies](#full-bodies)
+        - Storage: Not stored beyond a parameter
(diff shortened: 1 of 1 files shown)


src/specs/node_types.md
```diff
@@ -133,3 +133,12 @@ For convenience, we will define several common parameter configurations:
     - Fraud proofs:
         - State transition: produce, consume
         - Erasure coding: produce (partial), consume
+1. Validator nodes are identical to full nodes, but can additionally created consensus messages (proposals, votes, etc.).
```
liamsi, line 136:
```suggestion
1. Validator nodes are identical to full nodes, but should additionally create and sign consensus messages (proposals, votes, etc.).
```

adlerjohn:
Broke out removal of the DA header into https://github.com/celestiaorg/celestia-core/issues/866, so that this PR is self-contained to just fixing celestiaorg/celestia-specs#180.

liamsi (approved):
Makes sense. LGTM

PR #192, opened 2021-08-02
Update types.proto

nusret1996:
Added protobuf definitions for the fraud proofs and the data types used by them.

### DIFF
--- a/src/specs/proto/types.proto
+++ b/src/specs/proto/types.proto
@@ -37,3 +37,98 @@ message AvailableDataHeader {
   // array of 32-byte hashes
   repeated bytes col_roots = 2;
 }
+
+// https://github.com/celestiaorg/celestia-specs/blob/master/specs/data_structures.md#namespacemerkletreeinclusionproof
+message NamespaceMerkleTreeInclusionProof {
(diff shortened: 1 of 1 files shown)


nusret1996:
When creating a protobuf definition for the StateElement, shall I create a protobuf definition for individual state elements, e.g, Account, Delegation etc.? If so, how can I capture all of them when defining the type of the value within the StateElement? For references, [this](https://github.com/celestiaorg/celestia-specs/blob/ad4f7aaa9ef89d52da50826edcfe5d516a79e857/src/specs/data_structures.md#stateelement) is where StateElement is defined.

adlerjohn:
> shall I create a protobuf definition for individual state elements, e.g, Account, Delegation etc.?

Yes, otherwise no one can use the higher-level structures.

> If so, how can I capture all of them when defining the type of the value within the StateElement?

You can use `oneof`: https://developers.google.com/protocol-buffers/docs/proto3#oneof

liamsi:
Why was this closed?

adlerjohn:
https://github.com/celestiaorg/celestia-specs/pull/190#issuecomment-891823621

PR #207, opened 2022-03-28
Update link to celestia-app

rootulp:
The previous link redirects to https://github.com/celestiaorg/celestia-app but I think this change will reduce confusion


evan-forbes (approved):
:+1: thanks for this!

adlerjohn (approved):

PR #208, opened 2022-05-19
Docs: Update Sparse Merkle Tree base case specification to use default value

bvrooman:
Related issue: 
- https://github.com/celestiaorg/smt/issues/67

As per the discussion in the linked issue, the agreed upon change is that the root of an empty Sparse Merkle tree (the base case) is the _default value_, rather than the hash of the empty string. The default value, as per the spec, is the 32 byte 0: `0x0000000000000000000000000000000000000000000000000000000000000000`. This PR updates the specification to reflect the correct root in the SMT specification, and provides a short explanation on why this differs from the empty string root.

### DIFF
--- a/src/specs/data_structures.md
+++ b/src/specs/data_structures.md
@@ -377,11 +377,13 @@ Nodes contain a single field:
 |------|---------------------------|-------------|
 | `v`  | [HashDigest](#hashdigest) | Node value. |
 
-The base case (an empty tree) is defined as the [hash](#hashing) of the empty string:
+The base case (an empty tree) is defined as the default value:
 
 ```C++
-node.v = 0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+node.v = 0x0000000000000000000000000000000000000000000000000000000000000000
 ```
+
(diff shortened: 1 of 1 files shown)


src/specs/data_structures.md
```diff
@@ -377,12 +377,14 @@ Nodes contain a single field:
 |------|---------------------------|-------------|
 | `v`  | [HashDigest](#hashdigest) | Node value. |
 
-The base case (an empty tree) is defined as the [hash](#hashing) of the empty string:
+The base case (an empty tree) is defined as the default value:
 
 ```C++
-node.v = 0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
+node.v = 0x0000000000000000000000000000000000000000000000000000000000000000
```
adlerjohn, line 383: The hash of nothing still needs to be defined as the base case, as a user can configure an SMT with 0 leaves, as before. The root of a tree with only default leaves should separately be defined, as here.

bvrooman: Ah yes, I see - I'll separate these cases.
Is the implication that the base case is always an SMT of height 0? Is an unmodified SMT with height N not considered to be the base case?

adlerjohn: I wouldn't consider an empty SMT with more than zero leaves the vase case, I'd consider it like an initial case or default case.

adlerjohn (dismissed):

bvrooman:
I modified the wording slightly since your approval @adlerjohn and I think this better reflects the nomenclature discussion above. If this looks good to you, I just need a reapproval and for you to rerun the workflow.

adlerjohn (approved):

